PR #7, opened 2018-08-23
f4sefix from @hdmap

ghost:
This fixes https://github.com/ValveSoftware/Proton/issues/170

### DIFF
--- a/dlls/ntdll/virtual.c
+++ b/dlls/ntdll/virtual.c
@@ -1387,11 +1387,11 @@ static NTSTATUS map_image( HANDLE hmapping, ACCESS_MASK access, int fd, SIZE_T m
     server_enter_uninterrupted_section( &csVirtual, &sigset );
 
     if (base >= (char *)address_space_start)  /* make sure the DOS area remains free */
-        status = map_view( &view, base, total_size, mask, FALSE, SEC_IMAGE | SEC_FILE |
+        status = map_view( &view, base, total_size, mask, TRUE, SEC_IMAGE | SEC_FILE |
                            VPROT_COMMITTED | VPROT_READ | VPROT_EXEC | VPROT_WRITECOPY );
 
     if (status != STATUS_SUCCESS)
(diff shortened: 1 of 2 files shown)


Rabcor:
Somewhat disappointed this hasn't been applied yet after almost 2 months; I wonder why.

cjwijtmans:
maybe because the patches could introduce other problems but i think its worth putting it in a beta and keeping an eye on it.

Rabcor:
Well, there's no reason to think it will break anything else, but there's also no reason thinking it wouldn't. For all we know it might fix a bunch of other things unexpectedly with no downsides, who knows how many games would benefit from being able to do, well, these things https://github.com/hdmap/wine-hackery/tree/master/f4se . It's exactly why we have a beta right?

Personally I have a suspicion that anything negatively affected would be legacy software from the pre-XP era of windows.


I wonder though, if there's a possibility they could make it possible to create unique wine environments with these kinds of fixes only applied on a per-game basis, so that if oblivion/skyrim/skyrimse/fallout3/falloutnv/fallout4 are installed, it would use one patched version of wine but use an unpatched version for other ones? I suppose the tricky part might be doing it  without using too much space as those sorts of things might add up a bit eventually.

In this case for example, all it would probably require is a patched ntdll.dll; it shouldn't be that hard to store this patched ntdll.dll somewhere in proton and use it only in prefixes for games that need it.

ghost:
Hi!
This may sound stupid, but I couldn't find anywhere where your patch "Patch of Exile partial fix" would help this issue.
Is this patch intended, or an accidental addition?

kakra:
Sorry for the noise here, I changed the commit message by @zfigura to reference the correct repo now.

hdmap:
The get_free_mem_state_callback patch is merged into Wine as of 4.5 and the other patch is only required for older versions of f4se/skse64. After Protons Wine is synced to mainline Wine again this should be fixed.

PR #8, opened 2018-08-23
Fix for Path of Exile

ghost:
Partially fixes https://github.com/ValveSoftware/Proton/issues/58

### DIFF
--- a/dlls/ole32/compobj.c
+++ b/dlls/ole32/compobj.c
@@ -3237,6 +3237,7 @@ HRESULT WINAPI DECLSPEC_HOTPATCH CoCreateInstance(
 {
     MULTI_QI multi_qi = { iid };
     HRESULT hres;
+    CoInitialize(NULL);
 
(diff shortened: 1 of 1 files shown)


GloriousEggroll:
I think the patch by DataBeaver in the PoE issue on the proton repo is a little better/safer to use as it handles the call more gracefully:

https://gist.github.com/DataBeaver/5e1f1a256c7f1abb339831fc2b72a5db

MrSapps:
The caller should have called CoInitialize before CoCreateInstance.. seems unlikely this is "correct" to me..

cjwijtmans:
> The caller should have called CoInitialize before CoCreateInstance.. seems unlikely this is "correct" to me..

nowhere do i read from microsoft docs that CoInitialize must be called before CoCreateInstance, unless i am misunderstanding the docs since i dont know enough about ole.

aeikum:
Hello, this was fixed in Proton with e2ac9b7e3, a27b3269eb9, and 7af29148112. Two of those were upstreamed, one didn't make it in.

PR #12, opened 2018-08-28
Add support for some 3:2 resolutions

laloch:
See the comments under fce451cc6c064f58864139c9fd5fa95962d18b89.


aeikum:
Hey there, I added two resolutions in b8ad51545af8d2998a3ad264574d2b6337f53567, should hopefully be enough. I'd like to keep the list from growing too large.

tingvarsson:
Just saw this when browsing the wine uplift for proton, should 16:9 4k be added to the list as well? Is at least my native res. 3840x2160.
/Thanks

laloch:
@tingvarsson The native resolution should be picked up automatically. These are only "fallbacks".

@aeikum I didn't notice b8ad51545af8d2998a3ad264574d2b6337f53567. Would you want to play anything @960x640? I doubt it.
Edit: And you removed 1680x1050? That's pretty much the only resolution I use for gaming on my external 16:10 display.

aeikum:
@laloch I figure you play at the highest resolution your hw is capable of supporting. I tried to provide a range of settings for each ratio, each roughly 300px apart, rather than a few clustered together, to give a wide range of options depending on your hw power.

laloch:
OK, I see. Then please at least put back the 1680x1050. It was the only 16:10 resolution near 1024p.

aeikum:
@laloch Just curious, what's the native resolution of that display?

laloch:
It's 1920x1200 - the most common 16:10 display I guess.

aeikum:
Huh. I'm a little surprised the reduction to 1050 is enough to make a difference vs native. Anyway, I'll replace one of the other 8:5 resolutions with the 1050 one.

laloch:
Thank you. With the Intel graphics every pixel counts ;-)

tingvarsson:
@laloch OK then that should be fine yes, thanks for the quick explanation.

PR #15, opened 2018-09-02
crypt32: Strictly honor NULL parm in CryptBinaryToStringW

ryao:
This fixes the segfault issue in ValveSoftware/Proton#298. I need to wait for the wine-devel mailing list administrator to approve my request to join the mailing list before I can send the patch upstream and I wnat to get this fixed in Proton sooner, rather than later, so I am submitting it here first.

### DIFF
--- a/dlls/crypt32/base64.c
+++ b/dlls/crypt32/base64.c
@@ -304,13 +304,16 @@ static LONG encodeBase64W(const BYTE *in_buf, int in_len, LPCWSTR sep,
     needed = bytes + pad_bytes + 1;
     needed += (needed / 64 + 1) * strlenW(sep);
 
-    if (needed > *out_len)
-    {
+
+    if (out_buf == NULL) {
         *out_len = needed;
-        return ERROR_INSUFFICIENT_BUFFER;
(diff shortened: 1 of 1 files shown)


nsivov:
As already discussed in https://bugs.winehq.org/show_bug.cgi?id=44405 this is not a correct fix, is only pointing at the crash location.

ryao:
@nsivov I have revised this in a way that should be more correct, although there are a few loose ends:

1. No test cases yet.
2. I am still unable to attach a debugger to see what is happening here, so my understanding of the control flow is incomplete.

What is certain right now is that the revised version fixes a real problem in the current implementation where it will perform a comparison against potentially uninitialized data when a NULL parameter is set. Furthermore, this code should have done early exit when passed NULL from the start.

Is this okay for Proton in the short term or will it block on the full fix at upstream Wine? If it is the latter, one of us should close this. I can always reopen it when the full fix is ready in upstream Wine.

ryao:
I am withdrawing this due to a lack of time. :/

nsivov:
No need to close, if it's still an issue. I can take a look at adding tests for this.

PR #16, opened 2018-09-03
apply wininet-InternetCrackUrlW from wine-staging

GloriousEggroll:
Warframe cannot reach it's content update servers without this patch. Both the launcher and the game itself need this patch. The game will crash without this on launching due to not being able to access content servers. Fixes bug: [40598] Resize buffer when call to InternetCanonicalizeUrlW fails in InternetCrackUrlW on winehq


GloriousEggroll:
note: this patch was accepted into wine this morning. came through the dev mailing list

GloriousEggroll:
closing this as its now included in the 3.16-beta

PR #23, opened 2018-10-04
[fshack] winex11.drv: Bypass compositor in fshack mode

kakra:
When fshack is enabled, bypass the compositor in fullscreen mode. This
reduces stutter introduced by window updates in the background and also
allows for maybe a few more FPS.

I'm totally new to this stuff, please review and comment.


aeikum:
Can you be more specific about what problem this solves?

kakra:
It just improves performance for me, it doesn't solve a blocking problem. In Shadow of the Tomb Raider (in-game benchmark) this raised max-fps by about 5 fps (~10%) and avg-fps by about 1-2 fps. I thought it would be worth suggesting this patch.

It also solves some stutters I'm seeing when playing on the FHD-TV screen, and on the second monitor (4k) there are content updates to windows, i.e. a running htop in a terminal would introduce a very visible stutter per screen update.

This patch is probably a no-op in SteamOS if the window manager is configured to not do compositing. On a general desktop, tho, this patch should help performance without requiring users to explicitly tweak their settings.

aeikum:
What window manager are you using?

kakra:
I'm using kwin_x11 with NVIDIA proprietary driver.

kakra:
I explicitly decided to only set this in fshack mode to not disturb the user experience with compositing when a game is run in windowed mode. In my tests, games now successfully turn off compositing when the game switches to the fullscreen window, and if the game quits, the compositing state is restored. According to docs, the `NET_WM_BYPASS_COMPOSITING` is only a hint that the compositor is not forced to respect. The value 1 I'm passing to the X server says "suggest enabled" (2 would be "suggest disabled", 0 means "don't care").

kakra:
I've also researched a bit (mostly because to see how to actually do it), and I found patches for media players and SDL-fullscreen that also enable this WM hint. I din't check, tho, if these patches were actually applied. The idea of implementing this resulted from the subjective observation that games run smoother when kwin compositing was disabled here (no matter if I also saw stutters or not, that this also eliminates the stutters was a nice side-effect I only observed then during testing).

doitsujin:
Compositors in general tend to cause severe stutter issues when games are GPU-bound, so if this enables SDL-like behaviour on common compositors, it should definitely be considered in my opinion.

kakra:
Since I'm totally new to work within the wine code with self-made contributions, and with directly communicating with the X11 server in general, I recommend to first check if this hack is applying its function at the right location. But I tested it for some days now with multiple games and it works fine for me.

kakra:
@doitsujin I've also added a hack to my own dxvk build to enable triple-buffering by default (or how Vulkan calls it: mailbox mode with 2 back buffers). As soon as I disable vsync in games now (which seems to overwrite these hacked DXVK defaults), games run ultra-smooth now (but with lower FPS which settles at a divisor of 60, thus 20 or 30 here). But since this changes defaults, I probably don't want to suggest it for DXVK mainline.

This was somewhat unexpected because I thought that v-sync would lock the frame rate to a divisor of the refresh rate, and triple-buffering would work around that and allow for higher fps. I expected triple-buffering to introduce varying/alternating frame times (16.6, 33.3) with the benefit of reduced frame queue lag. Thus, I'm not sure what's going on here. Maybe the nvidia driver somehow interferes here (I've set it to do triple buffering by default, too, so maybe the driver opted out of this by disabling vsync in the game).

kakra:
I've reworked the patch to fix a minor code style issue, and also added the compositor note from @doitsujin  to the commit message.

aeikum:
Thanks for the review @doitsujin and the explanation @kakra. I'll look at pulling this in.

kakra:
I see this has been upstreamed in wine-proton 3.16. Closing.

zzhiyi:
I need to mention that this also introduce tearing.

kakra:
@zzhiyi I pretend you are referencing the original idea, not the "triple-buffering" patch.

I don't think that it **introduces** tearing, it's the other way around: The compositor takes countermeasures to work around tearing. If you bypass it, it's back to "normal" behavior.

Problems start when games use v-sync (or similar technologies to eliminate tearing) and the compositor is not bypassed. This may result in "double v-sync" and uneven frame timings, that would explain the stutter I was seeing without this patch. Flip v-sync on in the game, and it no longer tears.

But I think the most correct behavior is letting games handle this directly. Only with no compositor in the way we can have low latency frames for users who prefer low latency over tearing.

About the triple buffering patch: I'm probably going to remove it from my branch anyways, it's not working the way it should with current DXVK and/or NVIDIA. It's only linked here because I document my commits with links.

PR #28, opened 2018-10-20
Fix fullscreen mouse clipping

AlexeyProkhin:
A lot of users, including me, has an issue with a "mouse wall": if you move a mouse in one direction the camera eventually stops moving [[1]](https://github.com/ValveSoftware/Proton/issues/1418#issuecomment-420931532), [[2]](https://github.com/ValveSoftware/Proton/issues/832#issuecomment-420843944), [[3]](https://github.com/ValveSoftware/Proton/issues/579#issue-354033647). This happens because the pointer is not always successfully clipped in the fullscreen mode.

An important thing to understand is that [wine clips fullscreen windows automatically](https://github.com/ValveSoftware/wine/commit/a59c7cc8596b658d3c0854a737528cf60947ac1f) and, if the automatic procedure fails, an application will not be able to clip the mouse to the whole fullscreen window by calling ClipCursor.

The first three patches are fixes for the automatic fullscreen window clipping:
* If the screen mode changes rapidly, the mouse is not clipped
* The fs hack may break the fullscreen window detection when the resolution changes
* Unlike dxvk, wined3d does not trigger [a desktop resize event](https://github.com/ValveSoftware/wine/blob/58025614384053e61f1983a90f81135288f7e08e/dlls/winex11.drv/desktop.c#L323) when playing at the native resolution 

Those three patches alone fix the mouse clipping for me. I also include another one that changes ClipCursor behaviour so it activates the fullscreen clipping even if the automatic procedure has failed for some reason. I am not sure why it does not do it in the first place but I think it may be related to [this](https://github.com/ValveSoftware/wine/blob/58025614384053e61f1983a90f81135288f7e08e/dlls/user32/tests/monitor.c#L291). It makes sense to break windows compatibility here (especially with a bug!).

### DIFF
--- a/dlls/winex11.drv/mouse.c
+++ b/dlls/winex11.drv/mouse.c
@@ -1539,12 +1539,12 @@ BOOL CDECL X11DRV_ClipCursor( LPCRECT clip )
             }
             else if (grab_clipping_window( clip )) return TRUE;
         }
-        else /* if currently clipping, check if we should switch to fullscreen clipping */
+        else /* check if we should switch to fullscreen clipping */
         {
             struct x11drv_thread_data *data = x11drv_thread_data();
-            if (data && data->clip_hwnd)
+            if (data)
             {
-                if (EqualRect( clip, &clip_rect ) || clip_fullscreen_window( foreground, TRUE ))
+                if ((data->clip_hwnd && EqualRect( clip, &clip_rect )) || clip_fullscreen_window( foreground, TRUE ))
                     return TRUE;
             }
         }
--- a/dlls/winex11.drv/mouse.c
+++ b/dlls/winex11.drv/mouse.c
@@ -551,8 +551,10 @@ BOOL clip_fullscreen_window( HWND hwnd, BOOL reset )
     release_win_data( data );
     if (!fullscreen) return FALSE;
     if (!(thread_data = x11drv_thread_data())) return FALSE;
-    if (GetTickCount() - thread_data->clip_reset < 1000) return FALSE;
-    if (!reset && clipping_cursor && thread_data->clip_hwnd) return FALSE;  /* already clipping */
+    if (!reset) {
+        if (GetTickCount() - thread_data->clip_reset < 1000) return FALSE;
+        if (clipping_cursor && thread_data->clip_hwnd) return FALSE;  /* already clipping */
+    }
     rect = get_primary_monitor_rect();
     if (!grab_fullscreen)
     {


AlexeyProkhin:
An important thing to understand is that [wine clips fullscreen windows automatically](https://github.com/ValveSoftware/wine/commit/a59c7cc8596b658d3c0854a737528cf60947ac1f) and, if the automatic procedure fails, an application will not be able to clip the mouse to the whole fullscreen window by calling ClipCursor.

The first three patches are fixes for the automatic fullscreen window clipping:
* If the screen mode changes rapidly, the mouse is not clipped
* The fs hack may break the fullscreen window detection when the resolution changes
* Unlike dxvk, wined3d does not trigger [a desktop resize event](https://github.com/ValveSoftware/wine/blob/58025614384053e61f1983a90f81135288f7e08e/dlls/winex11.drv/desktop.c#L323) when playing at the native resolution 

Those three patches alone fix the mouse clipping for me. I also include another one that changes ClipCursor behaviour so it activates the fullscreen clipping even if the automatic procedure has failed for some reason. I am not sure why it does not do it in the first place but I think it may be related to [this](https://github.com/ValveSoftware/wine/blob/58025614384053e61f1983a90f81135288f7e08e/dlls/user32/tests/monitor.c#L291). It makes sense to break windows compatibility here (especially with a bug!).

dlls/winex11.drv/mouse.c
```diff
@@ -551,8 +551,10 @@ BOOL clip_fullscreen_window( HWND hwnd, BOOL reset )
     release_win_data( data );
     if (!fullscreen) return FALSE;
     if (!(thread_data = x11drv_thread_data())) return FALSE;
-    if (GetTickCount() - thread_data->clip_reset < 1000) return FALSE;
-    if (!reset && clipping_cursor && thread_data->clip_hwnd) return FALSE;  /* already clipping */
+    if (!reset) {
+        if (GetTickCount() - thread_data->clip_reset < 1000) return FALSE;
```
AlexeyProkhin, line 555: I do not understand why clip_reset is needed here. It probably should be just removed.

dlls/winex11.drv/mouse.c
```diff
@@ -547,7 +547,7 @@ BOOL clip_fullscreen_window( HWND hwnd, BOOL reset )
     /* maximized windows don't count as full screen */
     if ((style & WS_MAXIMIZE) && (style & WS_CAPTION) == WS_CAPTION) return FALSE;
     if (!(data = get_win_data( hwnd ))) return FALSE;
-    fullscreen = is_window_rect_fullscreen( &data->whole_rect );
+    fullscreen = fs_hack_enabled() || is_window_rect_fullscreen( &data->whole_rect );
```
AlexeyProkhin: Did I made correct assumption that fs_hack_enabled is true only when in the fullscreen mode?

AlexeyProkhin (commented):
41eb5418d451933608b0c7f10a7259024ac13d78 can be not applied if b0ad45c47a167fb149f9f12bb1b24e0edaaaeb62 would be. Actually, the opposite is also true.

aeikum:
@AlexeyProkhin Thanks very much for working on this. Since these changes will affect every game running in Proton, we will have to review and test these carefully, so it may take some time. To help verify the fix, can you let me know specifically what games and what window manager you were testing with when developing this?

aeikum:
@AlexeyProkhin Thanks very much for working on this. Since every game run through Proton will be affected by this, we'll have to review and test it carefully, which will take some time. To verify the fix, can you let me know what games and window manager you were testing with when developing these patches?

aeikum:
@AlexeyProkhin Thanks very much for working on this. Since every game run through Proton will be affected by this, we'll have to review and test it carefully, which will take some time. To verify the fix, can you let me know what games and window manager you were testing with when developing these patches?

(Sorry if you're getting a bunch of notes about this, either I'm very confused or something weird is going on with Github hiding my comments here.)

AlexeyProkhin:
@aeikum I have uploaded my system info to [gist](https://gist.github.com/AlexeyProkhin/1b8db28506f13872a8ddfacc046fb127). I tested on kubuntu 18.10, it was a beta at the time, but the issue was also present on kubuntu 18.04. 

The changes are not tested very well. I mostly tested it with Witcher 3 and later confirmed that the patches are working on Gta V.
 
So yes, the patches are very invasive and I expected they will not be merged quickly (or at all). Feel free to cherrypick from them or base your work upon them. At a very least, they show the relevant places where the clipping issue occur.

P.S. GitHub has some database issues today, so you may get error 404 when opening it. Just keep refreshing until it open.

dlls/winex11.drv/mouse.c
aeikum: This looks wrong to me. `is_window_rect_fullscreen` should already handle the fullscreen hack via the primary monitor getting resized in `xinerama_init`. `fs_hack_enabled` only indicates that we are emulating a mode change for the application, not that there is a fullscreen window present.

dlls/winex11.drv/mouse.c
aeikum: This change makes sense to me and could maybe go upstream.

dlls/winex11.drv/mouse.c
aeikum: Do you recall what this change fixed? Maybe something with multiple monitors? Maybe we can fix it another way.

dlls/winex11.drv/mouse.c
AlexeyProkhin: The only thing I don't like about it is that I make the change specifically to fix cases when the resolution is changed, i.e. when clip_fullscreen_window() is called from X11DRV_resize_desktop(). But reset is also true when clip_fullscreen_window() is called X11DRV_ClipCursor(). It should not matter with the vanilla wine because it never called if the mouse is not already clipped, but it would matter if b0ad45c47a167fb149f9f12bb1b24e0edaaaeb62 is applied.

But like I said, I don't understand why clip_reset is here, so I cannot judge when we can safely ignore it.

dlls/winex11.drv/mouse.c
AlexeyProkhin: > fs_hack_enabled only indicates that we are emulating a mode change for the application, not that there is a fullscreen window present.

Yeah... the change is probably wrong. It attempts to fix two issues on my machine:
1. I don't remember exactly, but I think when the game is started and the fs hack is enabled, data->whole_rect is a rect that has the same size as the primary screen but slightly moved to one side. I will look into it again tomorrow and give more info.
2. When changing resolution, data->whole_rect may have the size same as the old resolution and not the new one.

I have tested it with one monitor. I have a tv connected to the video card but it was disabled in the settings. It may have interfered. I also will look into it.

AlexeyProkhin: Come to think of it, the issues probably do occur elsewhere. clip_fullscreen_window() seems to be called a bit too soon before the window size is updated.

aeikum:
OK after looking over this, I think b0ad45c47a167fb149f9f12bb1b24e0edaaaeb62 and 70b789a86ea53cefee71aa471dfd3cdd302d56f2 make sense to me. You mentioned that we don't need the wined3d commit in that case. That leaves f577e79c43661b5e4d25cd6596cfc76faa94cd5e, which we are discussing.

AlexeyProkhin:
Good news! I just checked and b0ad45c47a167fb149f9f12bb1b24e0edaaaeb62 and 70b789a86ea53cefee71aa471dfd3cdd302d56f2 are enough to fix the mouse clipping for me at least with Witcher 3 and Gta V.
clip_fullscreen_window() called from X11DRV_resize_desktop() fails to grab the cursor however. I think it is because the fs hack calls XMoveResizeWindow() and whole_rect is not yet updated when clip_fullscreen_window() is called.

aeikum:
> clip_fullscreen_window() called from X11DRV_resize_desktop() fails to grab the cursor however. I think it is because the fs hack calls XMoveResizeWindow() and whole_rect is not yet updated when clip_fullscreen_window() is called.

Does this show up as a problem clipping the cursor in a game?

AlexeyProkhin:
> > clip_fullscreen_window() called from X11DRV_resize_desktop() fails to grab the cursor however. I think it is because the fs hack calls XMoveResizeWindow() and whole_rect is not yet updated when clip_fullscreen_window() is called.
> 
> Does this show up as a problem clipping the cursor in a game?

No! The clipping works as expected. b0ad45c47a167fb149f9f12bb1b24e0edaaaeb62 takes care of it. This is what happens:
1. Games changes the resolution
2. X11DRV_resize_desktop is eventually called
3. It calls clip_fullscreen_window() which fails
4. A bit later game calls ClipCursor()
5. clip_fullscreen_window() is called again. This time, thanks to b0ad45c47a167fb149f9f12bb1b24e0edaaaeb62, the clipping is successfully applied.

AlexeyProkhin:
I just have removed wined3d patch and f577e79c43661b5e4d25cd6596cfc76faa94cd5e from the pull request. The two commits that are left should fix any clipping issue. And if later for some DE/game it would not be the case, well, at least we would know the first place to look into.

aeikum:
Thanks! I've asked our QA team to test with your changes applied, both to confirm it's fixed in the games you've mentioned here, and to try to detect any new failures. I'll keep you posted on how it goes.

aeikum:
@AlexeyProkhin, those two commits were pulled into 3.16-4. If you feel this PR is complete, please close it. If you've got more work in this area, let us know! Thanks again for looking into this.

PR #29, opened 2018-10-25
[DO NOT MERGE] Rewrite XAudio libraries to use FAudio

flibitijibibo:
This isn't a pull request as much as it is a quick way to track the FAudio changes for Wine. This is based on a WIP proton_3.7 branch so it wouldn't merge even if you wanted to...

My TODO List:
- [x] Use CustomAllocatorEXT extension
- [x] Fix `xaudio2_x` builds
- [x] Rewrite `x3daudio.c`
- [x] Rework XAPO implementation
- [x] Write `xactengine3_x`
- [x] Dynamically link to FAudio, don't statically link
- [x] Merge XAudio2 files back into monolithic file
- [x] Merge XACT3 files into monolithic file


aeikum:
It's not a deal-killer, but I don't see the advantage to splitting this into a bunch of files. It kills our git history, and makes finding the code you're looking for a chore. E.g. say there's a change to the voice code, now I need to open three source files instead of just one.

Also just want to make sure you're planning for some way for audio to go through Wine's audio subsystems at some point. Maybe that's part of the last checkbox.

flibitijibibo:
Funnily enough I split up the files for readability purposes, but now that I've gone through all the sources I should be able to bring it all back into one file for the final patch. (My XACT stuff is separate as well but it should be much easier to bundle that into one file after it's done.)

The last checkbox depends mostly on what the Wine team decides on doing; I'm still up for the dynamic link since CustomAllocatorEXT basically fixes this but I'm also up for keeping FAudio in-tree for a WASAPI backend. Exposing the platform stuff is still off the table since I'm not 100% sure if that's something I'd be willing to stabilize on for all eternity when it's still getting changed as recently as a few days ago, and I'm also still not sure what the team wants to happen when packagers inevitably don't build with FFmpeg support.

flibitijibibo:
The XACT DLLs should now compile. I'm having a bit of trouble with the .idl stuff, but it's all pointer types so "technically" it's not a huge deal. They definitely need to be cleaned up to kill all the warnings though (grep `FIXME:` in xact3.idl). Also need to deal with unnamed unions but it shouldn't affect runtime compatibility.

flibitijibibo:
Just wrapped up the last of the tangible changes, at this point the code should be complete and usable. There are still a couple fussy warnings in xact3.idl, and there's one gap in FAudio in that there's no real way to "install" it, I've just been throwing the headers and libFAudio.so in my system's /usr/local/. Somehow I've gone nearly 7 years without ever making a library that's supposed to be installed... EDIT: Added some basic install/uninstall targets, so that's been dealt with now.

The last thing that needs to be done is merging all the files back together into the big `_dll.c` files, but that's entirely cosmetic.

flibitijibibo:
Just finished merging all the files back together and everything still works, so I _think_ the initial reviewing can start now. @aeikum, if you don't see anything immediately objectionable I'll redo this patch against whatever branch you think is best (proton_3.16, or something else?).

aeikum:
Thanks @flibitijibibo. I'll try to review this soon. I think it makes sense to apply it on top of current winehq master. That code doesn't change very much, so I imagine  rebase would be mostly trivial. I'll start by testing it in upstream Wine with my usual suite of xaudio games.

Have you run this build against Wine's xuadio2 tests?

It's not what I would've chosen, but I think I can live with outputting from the library to SDL for now. If FAudio settles down in the future, or we run across some problem that can't be solved without it, we can look at making the internal API available.

flibitijibibo:
I haven't just yet, I've only looked at the source to it - it seemed pretty much like what the XACT runtime does, just not in a unit test form. The only test I am unsure about is the CreateFX test, which seems to call CreateFX with the AudioReverb/AudioVolumeMeter CLSIDs, is that valid in 2.7 and under? I can see it working in 2.8+, but I wasn't sure if that somehow worked despite the MSDN docs saying otherwise.

Once the subject of output mixing comes up again I'll try to put out another idea... I'm wondering if it would be more helpful to look at this as an even-lower-level problem, since doing an internal API override in FAudio would only solve it for XAudio2, when we also have openal32, dsound, etc. to worry about as well. One crazy idea I had was adding support for WASAPI in the Linux-native version of SDL2, but that'd still have to be enabled in OpenAL (Soft 1.19 has an SDL backend!) and whatever else is used for audio support.

flibitijibibo:
Rebased and compressed against WineHQ master, diff can be viewed here:

https://github.com/flibitijibibo/wine/pull/1

flibitijibibo:
Closed in favor of https://github.com/ValveSoftware/wine/pull/32

PR #30, opened 2018-10-28
Fix typo in msi

simons-public:
This fixes a regression in wine that was [identified in the mailing list](https://www.winehq.org/pipermail/wine-cvs/2018-September/130002.html). It prevents the installation of dotnet40 in winetricks, and probably other software that uses msi to install.

The fix has already been committed to upstream wine.


aeikum:
Thanks, will pick that commit across.

aeikum:
This was picked as e2419cdf58af7da12a9b886c334c1a0ccd862b14, thanks for the suggestion.

PR #33, opened 2018-11-06
Fixes for Elite Dangerous on Proton 3.16

redmcg:
This pull request includes various patches placed on Proton version 3.16-4 Beta that fixes the following issues with Elite Dangerous: 
- Crash during opening video (https://github.com/ValveSoftware/wine/commit/24f8b0dec37eab525a276a44c8e7a37e23d2d425)
- CRC Error when logging on to Frontier servers (https://github.com/ValveSoftware/wine/commit/09dd4dc858263c6c5706e93571c7d929b6a43fa2)
- Freezing after a few minutes of game-play (https://github.com/ValveSoftware/wine/commit/30a0138264e0e6ddd6ae0f53e6dc75280a400526)
- Keybind pre-sets not showing up (https://github.com/ValveSoftware/wine/commit/280821fb01882000476bcb71d4cae1ba58d0aebd)
- Keyboard not working correctly (https://github.com/wine-staging/wine-staging/tree/master/patches/kernel32-NormalizeString)

The CRC error is fixed by running the Launcher with 'wine64'. The included patch allows 32-bit .NET applications to run under wine64 - i.e. it fixes the error being worked around here: https://github.com/ValveSoftware/Proton/commit/97db8b5d8d2109a3c93e11f1d6f2e8e8957774a0 (thus this would need to be reverted)

All these patches have been accepted upstream in wine or wine-staging (with the exception of the fix used for the opening video crash - this is a variation of the one I had accepted to the 'wine' repo as it applied easier to 3.16).

The two that are currently only in wine-staging are:
1. the futex-based condition variable (which fixes the freezing after a few minutes of game-play); and
2. the NormalizeString API (which fixes the keyboard not working correctly)


aeikum:
Thanks for the suggestions, we'll take a look.

redmcg:
@aeikum Just an FYI - I found a bug that would cause problems using wine64 with .NET applications that have the 32-bit mandatory flag switched on. I've just submitted a [fix](https://www.winehq.org/pipermail/wine-devel/2018-November/135802.html) to to the wine-devel mailing list.

redmcg:
This fix has now been accepted. See: https://github.com/wine-mirror/wine/commit/8cd569cf2628fbb2497e5ef020e670f77cf36d40

aeikum:
OK, we can pull in the ntdll changes and the dinput patch. I'd rather wait for feedback from upstream on the NormalizeString patches (thanks for trying to keep that moving). I'm also unsure about the futex one, I know Zebediah was planning to push it upstream himself. If it doesn't get into 4.0, I'll talk to him about it.

WRT to the quartz one, I'm undecided on whether it makes more sense to pull in all of the changes, or just apply your small diff.  It's tricky because quartz code is spread across a bunch of modules, so it's hard to be certain we got every change we need.

aeikum:
Oh, I'm also confused about what you mean about launching things with wine64. My understanding is everything should work with just the wine executable. After those ntdll patches are applied, does your case still require using wine64, and just plain wine does not work?

redmcg:
> I'm also unsure about the futex one, I know Zebediah was planning to push it upstream himself. If it doesn't get into 4.0, I'll talk to him about it.

Zebediah may be planning to change his approach here. We found a bug in the existing implementation that can be fixed using the RtlWaitOnAddress()/RtlWakeAddress() calls; which look to have been added end of October (after he made his futex patch). So I think he may now plan to use these functions and add the futex path to these.

You can track the latest here:
https://bugs.winehq.org/show_bug.cgi?id=46208

redmcg:
> Oh, I'm also confused about what you mean about launching things with wine64. My understanding is everything should work with just the wine executable.

The one exception to this is 32-bit Mono/.Net assemblies. If you run with 'wine', the interpreter will create 32-bit machine code and if you run with 'wine64' it will create 64-bit machine code.

With 32-bit machine code you'll link to 32-bit unmanaged DLLs and access the WoW6432Node of the registry.

This is usually fine, but EDLaunch (the Elite launcher) requires access to the 'MachineGuid' value stored under the 'HKLM\Software\Microsoft\Cryptography' key - which doesn't have a WoW6432Node entry.

> After those ntdll patches are applied, does your case still require using wine64, and just plain wine does not work?

So that's correct - The 'EDLaunch' executable needs to be executed with 'wine64'.

Although prior to https://github.com/ValveSoftware/wine/commit/09dd4dc858263c6c5706e93571c7d929b6a43fa2, the work-around was to copy the 'MachineGuid' value to the WoW6432Node (and then you could run with 'wine').

aeikum:
Hi, a few of these were merged into the 3.16-5 release: d98d6d0da344cfd26f82d9284e8e2622ffda748b, aa9d31039e751d563734b3e929e8da90638e22b4, a0a7702c06c73fb0a0a7b82ddf39fcb6433cb8a1.

redmcg:
Thanks Andrew. I'll close this pull request.

I'm guessing the other fixes will need to wait until Proton rebases on a Wine release that includes them?

SteveDevOps:
Will these patches be forthcoming in 3.16.5 release? I just tried launching ED on 3.16.4 (release) and 3.16.5 (beta) -- no luck. On Pop!_OS (ubuntu 18.04 fork).

redmcg:
@SteveDevOps If you haven't already - you might want to try my custom build that includes these patches:
https://github.com/redmcg/wine/releases/

re: 3.16.5 - I'll await to hear back from Andrew

PR #36, opened 2018-12-13
add necessary lib flags for faudio with ffmpeg

GloriousEggroll:
These need to be added in order for faudio compiled with ffmpeg to work with wine. This is to be used with this PR: https://github.com/ValveSoftware/Proton/pull/2082

### DIFF
--- a/configure
+++ b/configure
@@ -13895,7 +13895,7 @@ if ${FAUDIO_LIBS:+false} :; then :
 fi
 fi
 
-FAUDIO_LIBS=${FAUDIO_LIBS:-"-lFAudio"}
+FAUDIO_LIBS=${FAUDIO_LIBS:-"-lFAudio -lavcodec -lavutil -lswresample"}
 $as_echo "$as_me:${as_lineno-$LINENO}: faudio cflags: $FAUDIO_CFLAGS" >&5
 $as_echo "$as_me:${as_lineno-$LINENO}: faudio libs: $FAUDIO_LIBS" >&5
 ac_save_CPPFLAGS=$CPPFLAGS


kakra (changes requested):
`configure` is a generated file, it shouldn't be patched directly. `configure.ac` should be patched, then re-run `autoconf`.

GloriousEggroll:
@kakra the previous changes for faudio were made to configure. I followed what was already done.

https://github.com/ValveSoftware/wine/commit/967f095438dd487080216d5906f4b68994b1752b#diff-e2d5a00791bce9a01f99bc6fd613a39d

line 13890

kakra:
@GloriousEggroll No, not really... That referenced commit changes `configure.ac`, then `autoconf` was run and regenerated the `configure` file. Both changes were committed then. Don't let that fool you. ;-)

Also, following bad examples is not always wise. :-)

Upstream wine usually does it a little different: Patches to `configure` are usually not accepted except from the core devs who regenerate with `autoconf` and then commit the result (maybe melting it into the downstream commit). This saves a lot of headache during rebase because you end up with fewer merge conflicts due to changes in `configure`.

aeikum:
Those little nitpicks are things we can clean up at merge-time, they wouldn't block a PR.

aeikum:
Oh, I misunderstood and thought this included both configure.ac and configure changes. Yeah, @kakra  is right, this should be changed in configure.ac and the configure file regenerated from that.

kakra:
@aeikum Yeah, otherwise one could unknowingly loose changes to `configure`...

kakra:
I'm not sure if this is needed anyways... The default configure script uses pkgconfig to determine the linker flags. With system-wide ffmpeg this should work correctly by default because all libs for full decoder support should be found if system-wide ffmpeg supports them.

For the other case, that is a bundled ffmpeg, the build scripts should manually include the faudio linker flags. That is, the Makefile should add/override proper `FAUDIO_LIBS`. Otherwise you end up with mixed up dependencies during configure, build, and make time. Or does my understanding completely fail?

GloriousEggroll:
@kakra just a bit of background on how I came up with the proposed solutions: I've been working with this the last 3 weeks or so giving feedback to Ethan while testing Warframe with users who use my warframe launcher.

-When you specify --with-ffmpeg, it builds it's own version of ffmpeg, which faudio then builds against. If you don't specify to faudio where the FFmpeg_INCLUDE_DIRS folder is, which is obj-tools32/64, it builds against the system's ffmpeg, which is version .53 instead of version .58. WINE will complain about this when compiling. Once compiled, FAudio will not be able to locate the necessary ffmpeg libraries and the game using it will not run properly. An example of this is Warframe crashes right after the initial loading screen if you use version .53 instead of .58.

-Additionally, the flags are needed, otherwise WINE fails at testing for faudio with the following:
```
configure:13935: gcc -m64 -o conftest -I/home/vagrant/build/obj-tools64/include -g -O2 -march=nocona -mtune=core-avx2 -mfpmath=sse -fwrapv -fno-strict-aliasing   -L/home/vagrant/build/obj-tools64/lib conftest.c -lFAudio -lFAudio  >&5
/usr/bin/ld: warning: libavcodec.so.58, needed by /home/vagrant/build/obj-tools64/lib/libFAudio.so, not found (try using -rpath or -rpath-link)
/usr/bin/ld: warning: libavutil.so.56, needed by /home/vagrant/build/obj-tools64/lib/libFAudio.so, not found (try using -rpath or -rpath-link)
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `avcodec_receive_frame@LIBAVCODEC_58'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `avcodec_send_packet@LIBAVCODEC_58'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `avcodec_find_decoder@LIBAVCODEC_58'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `avcodec_alloc_context3@LIBAVCODEC_58'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `av_frame_alloc@LIBAVUTIL_56'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `avcodec_close@LIBAVCODEC_58'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `avcodec_open2@LIBAVCODEC_58'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `avcodec_flush_buffers@LIBAVCODEC_58'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `av_malloc@LIBAVUTIL_56'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `av_sample_fmt_is_planar@LIBAVUTIL_56'
/home/vagrant/build/obj-tools64/lib/libFAudio.so: undefined reference to `av_free@LIBAVUTIL_56'
```
-pkgconfig doesn't work for faudio, because there is no faudio.pc file for it to pull information from. I also did it through this method when I was testing by making an faudio.pc file get generated, but it was overkill when the same thing was achieved by just hard coding the flags. Here's how I generated it for testing for proton's build file, which also works:
```
faudio64: SHELL = $(CONTAINER_SHELL64)
faudio64: $(FAUDIO_CONFIGURE_FILES64)
	+$(MAKE) -C $(FAUDIO_OBJ64) VERBOSE=1
	+$(MAKE) -C $(FAUDIO_OBJ64) install VERBOSE=1
	mkdir -p $(DST_DIR)/lib64

	echo "prefix=$(abspath $(TOOLS_DIR64))" > $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc
	echo "libdir=\$${prefix}/lib" >> $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc
	echo "includedir=\$${prefix}/include" >> $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc
	echo "" >> $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc
	echo "Name: faudio" >> $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc
	echo "Description: Accuracy-focused XAudio reimplementation for open platforms" >> $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc
	echo "Version:" >> $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc
	echo "" >> $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc
	echo "Libs: -L\$${libdir} -lFAudio -lavcodec -lavutil -lSDL2 -lswresample" >> $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc
	echo "Cflags: -I\$${includedir}" >> $(TOOLS_DIR64)/lib/pkgconfig/faudio.pc

	cp -L $(TOOLS_DIR64)/lib/libFAudio* $(DST_DIR)/lib64/
	[ x"$(STRIP)" = x ] || $(STRIP) $(DST_DIR)/lib64/libFAudio.so

faudio32: SHELL = $(CONTAINER_SHELL32)
faudio32: $(FAUDIO_CONFIGURE_FILES32)
	+$(MAKE) -C $(FAUDIO_OBJ32) CFLAGS=-m32 VERBOSE=1
	+$(MAKE) -C $(FAUDIO_OBJ32) install VERBOSE=1
	mkdir -p $(DST_DIR)/lib

	echo "prefix=$(abspath $(TOOLS_DIR32))" > $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc
	echo "libdir=\$${prefix}/lib" >> $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc
	echo "includedir=\$${prefix}/include" >> $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc
	echo "" >> $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc
	echo "Name: faudio" >> $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc
	echo "Description: Accuracy-focused XAudio reimplementation for open platforms" >> $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc
	echo "Version:" >> $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc
	echo "" >> $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc
	echo "Libs: -L\$${libdir} -lFAudio -lavcodec -lavutil -lSDL2 -lswresample" >> $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc
	echo "Cflags: -I\$${includedir}" >> $(TOOLS_DIR32)/lib/pkgconfig/faudio.pc

	cp -L $(TOOLS_DIR32)/lib/libFAudio* $(DST_DIR)/lib/
	[ x"$(STRIP)" = x ] || $(STRIP) $(DST_DIR)/lib/libFAudio.so
```

I posted my method because in the last 3 weeks or so that I've been working on it, this is what worked. As far as I could tell it was more reliable to package avcodec, avutil, and swresample along with faudio than to rely on it trying to detect them and use whatever version was on a system, if at all, and if it doesnt detect them it fails miserably. 

Worst case scenario this PR and the one that goes with it can be used to pull information from when Valve decides to try integrating it, rather than going in the same circles I did.

kakra:
I'm currently using this solution in my own build system:
https://github.com/kakra/wine-proton/blob/rebase/proton_3.16/makefile

But first, my system has ffmpeg with wma support. And second, I don't own the game you're testing against, so I cannot cross-check. Any other games the would depend on this?

GloriousEggroll:
-Not all systems have ffmpeg installed, some users, especially new users, are completely unaware. As far as proton goes, the general idea is for it to run with the least amount of user "interaction" needed, for ease of use, right? For that to happen imo the necessary ffmpeg libraries need to be packaged with it. Just my two cents

-"An example of this is Warframe crashes right after the initial loading screen if you use version .53 instead of .58."

Also, warframe is free to play. It needs my launcher to be able to run, but it is a good candidate because there are many things wrong with it that faudio fixes.

My method of building is using vagrant + docker per Valve's README instructions

My warframe launcher and instructions if you feel like testing faudio with it:
https://gitlab.com/GloriousEggroll/warframe-linux/tree/steamplay-proton
https://gitlab.com/GloriousEggroll/warframe-linux/wikis/Installation-Instructions

You'll want to comment out lines 172-189 in Launcher.sh if testing with a custom proton build, as it manually applies the com wrapper versions of faudio and an older build of dxvk otherwise upon first run.

mat8913:
@GloriousEggroll I believe these errors you're getting indicate that FAudio was not linked to ffmpeg correctly. Could you please post the output of `objdump -p /home/vagrant/build/obj-tools64/lib/libFAudio.so`?

GloriousEggroll:
[objdump.txt](https://github.com/ValveSoftware/wine/files/2679202/objdump.txt)

and here is the faudio section of my modified makefile_base.mak for proton:
https://github.com/GloriousEggroll/Proton/blob/f555e5a7bda0dfed6f7bfe02d9aa3b0e0ab7e039/build/makefile_base.mak#L526

as mentioned - I only get those errors if I do not specify -lavcodec -lavutil -lswresample in the flags for faudio's libs per this PR's changes

kakra:
@GloriousEggroll Do you know if binary distribution of ffmpeg still has issues with various license holders?

mat8913:
@GloriousEggroll I did some more reading and I think I know what's wrong: `ld` doesn't search paths specified by `-L` when looking for second-order dependencies.

I haven't been able to test it yet, but this patch to makefile_base.mak for proton should fix it:

```patch
diff --git a/build/makefile_base.mak b/build/makefile_base.mak
index 77038c0..5e9fa6d 100644
--- a/build/makefile_base.mak
+++ b/build/makefile_base.mak
@@ -704,7 +704,7 @@ $(WINE_CONFIGURE_FILES64): $(MAKEFILE_DEP) | $(WINE_OBJ64)
 	cd $(dir $@) && \
 		STRIP=$(STRIP_QUOTED) \
 		CFLAGS=-I$(abspath $(TOOLS_DIR64))"/include -g $(COMMON_FLAGS)" \
-		LDFLAGS=-L$(abspath $(TOOLS_DIR64))/lib \
+		LDFLAGS="-L$(abspath $(TOOLS_DIR64))/lib -Wl,-rpath-link,$(abspath $(TOOLS_DIR64))/lib" \
 		PKG_CONFIG_PATH=$(abspath $(TOOLS_DIR64))/lib/pkgconfig \
 		CC=$(CC_QUOTED) \
 		CXX=$(CXX_QUOTED) \
@@ -719,7 +719,7 @@ $(WINE_CONFIGURE_FILES32): $(MAKEFILE_DEP) | $(WINE_OBJ32) $(WINE_ORDER_DEPS32)
 	cd $(dir $@) && \
 		STRIP=$(STRIP_QUOTED) \
 		CFLAGS=-I$(abspath $(TOOLS_DIR32))"/include -g $(COMMON_FLAGS)" \
-		LDFLAGS=-L$(abspath $(TOOLS_DIR32))/lib \
+		LDFLAGS="-L$(abspath $(TOOLS_DIR32))/lib -Wl,-rpath-link,$(abspath $(TOOLS_DIR32))/lib" \
 		PKG_CONFIG_PATH=$(abspath $(TOOLS_DIR32))/lib/pkgconfig \
 		CC=$(CC_QUOTED) \
 		CXX=$(CXX_QUOTED) \
```

If somebody is willing to test this, it would be much appreciated. I'm still waiting for `vagrant up` to finish.

kakra:
@mat8913 Yes, that's exactly the problem. But AFAIK using rpath does not help because it hard-codes the absolute library path into the linker parameters. It should be enough to put `-Wl,-lFAudio,-lavcodec,-lavutil,-lswresample`. I solved a similar problem here and was additionally experimenting with `LD_LIBRARY_PATH`. But all solutions seemed wrong and introduced other problems at the other end.

What matters is the order of build recipes. The dependencies must be present first. I solved this by configuring and building the dependencies first, and installing them to the dist directory, before configuring the depending subprojects. You can then give it the dist path in `-L` and things should start to work.

mat8913:
@kakra According to [this](https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_node/ld_3.html), rpath-link shouldn't do that.

That's a fairly long page, so I'll copy the relevant section here:

>  The difference between -rpath and -rpath-link is that directories specified by -rpath options are included in the executable and used at runtime, whereas the -rpath-link option is only effective at link time.

kakra:
@mat8913 Ah, okay... I wasn't aware of that being the difference. I only learned to avoid rpath by personal experience. ;-)

GloriousEggroll:
> @GloriousEggroll Do you know if binary distribution of ffmpeg still has issues with various license holders?

sorry I do not. 

@mat8913 I won't be able to test until later today. it was about 3:30 am when I got your reply and I'm unfortunately at work for the next 7 hours or so

mat8913:
I've got it building now, looks promising so far:
```
checking FAudio.h usability... yes
checking FAudio.h presence... no
configure: WARNING: FAudio.h: accepted by the compiler, rejected by the preprocessor!
configure: WARNING: FAudio.h: proceeding with the compiler's result
checking for FAudio.h... yes
checking for -lFAudio... libFAudio.so
```

@kakra I think I've encountered the build order problem. If I run `make dist` it tries to build wine before it's built faudio, however just running `make` seems to build in the correct order.

kakra:
@mat8913 Did you run `make configure` first? It should build `faudio` before even configuring wine...

**EDIT:** Thinking about this again, I'm not sure if you used my git repository... Ignore what I suggested if you didn't. ;-)

mat8913:
@kakra I meant with Valve's Makefile, not with your one. Their Makefile doesn't have a `configure` target:
```
vagrant@debian9:~/build$ make configure
make configure -f Makefile NO_NESTED_MAKE=1
make[1]: Entering directory '/home/vagrant/build'
make[1]: *** No rule to make target 'configure'.  Stop.
make[1]: Leaving directory '/home/vagrant/build'
../proton/build/makefile_base.mak:17: recipe for target 'nested_make' failed
make: *** [nested_make] Error 2
```

kakra:
@mat8913 Your findings look familiar, I think you're on the right track.

GloriousEggroll:
thanks @mat8913 that resolved the issue. closing this pr as it's no longer needed. resolved by https://github.com/GloriousEggroll/Proton/pull/1/commits/c2a2696cff7a6c1ad6bf3a1dd21b22d9eabee885

mat8913:
@kakra Yeah I wasn't using your git repo, sorry for the confusion. I just thought I might've been running into the same problem as you.

GloriousEggroll:
Re-opening due to new findings:

https://github.com/ValveSoftware/Proton/issues/3274

Relevant patch to make it work without rpath-link:

~~~
diff --git a/configure.ac b/configure.ac
index 5e36331596..8f4ad7627c 100644
--- a/configure.ac
+++ b/configure.ac
@@ -1741,7 +1741,7 @@ WINE_NOTICE_WITH(sdl,[test "x$ac_cv_lib_soname_SDL2" = "x"],
 dnl **** Check for FAudio ****
 if test "x$with_faudio" != "xno"
 then
-    WINE_PACKAGE_FLAGS(FAUDIO,[faudio],[-lFAudio],,,
+    WINE_PACKAGE_FLAGS(FAUDIO,[faudio],[-lFAudio -lavcodec -lavutil -lswresample],,,
         [AC_CHECK_HEADERS([FAudio.h],
             [WINE_CHECK_SONAME(FAudio,FAudioCreate,,,[$FAUDIO_LIBS],[[libFAudio*]])])
              WINE_CHECK_LIB_FUNCS([FAudio_CommitOperationSet \
diff --git a/configure b/configure
index 818d31d270..bb3c4b5fbd 100755
--- a/configure
+++ b/configure
@@ -13895,7 +13895,7 @@ if ${FAUDIO_LIBS:+false} :; then :
 fi
 fi
 
-FAUDIO_LIBS=${FAUDIO_LIBS:-"-lFAudio"}
+FAUDIO_LIBS=${FAUDIO_LIBS:-"-lFAudio -lavcodec -lavutil -lswresample"}
 $as_echo "$as_me:${as_lineno-$LINENO}: faudio cflags: $FAUDIO_CFLAGS" >&5
 $as_echo "$as_me:${as_lineno-$LINENO}: faudio libs: $FAUDIO_LIBS" >&5
 ac_save_CPPFLAGS=$CPPFLAGS
~~~

aeikum:
This doesn't look like the right solution to me. Our libraries don't link to avcodec, etc, so we shouldn't need to add them to the linker args. Why can't your libFAudio find its dependencies at runtime? `LD_DEBUG` might be helpful here.

pchome:
@GloriousEggroll 
Just generate necessary *.pc files for ffmpeg and faudio, place them into `$(abspath $(TOOLS_DIR64))/lib/pkgconfig` and `$(abspath $(TOOLS_DIR32))/lib/pkgconfig`.
Wine should pick them, since using `PKG_CONFIG_PATH=$(abspath $(TOOLS_DIR64))/lib/pkgconfig` environment variable.

Note: I'm not using virtual environment to build wine, so I'm not sure if this is the real root of the problem here. For this Wine built on real system with system ffmpeg and faudio everything seems works fine.

GloriousEggroll:
Closing this as I've tested this with wine and it is working again:
```
LDFLAGS="-L$(abspath $(TOOLS_DIR64))/lib -Wl,-rpath-link,$(abspath $(TOOLS_DIR64))/lib" \
```
additional if this method is not correct, pchome's method may also be a good approach.

PR #40, opened 2019-01-09
msacm32: Reference count local drivers.

zzhiyi:
One local driver can be assigned to multiple driver ids.
When releasing the driver id, check if the reference count
of a local driver is one before actually releasing the local driver.

Fix https://github.com/ValveSoftware/Proton/issues/210

Signed-off-by: Zhiyi Zhang <zzhang@codeweavers.com>


aeikum:
This was included in 3.16-7. df661f44e720c8d401acc69859a5f13b3efa8350

PR #47, opened 2019-04-26
winex11: Fill WM_CLASS based on Steam appid

dreamer:
Some desktop environments (Gnome 3, Cinnamon) decide on an application
icon in the following order:

- If the first string in WM_CLASS property can be correlated to
  a name or StartupWMClass key in a .desktop entry file, then
  the associated icon will be used.
- If the second string in WM_CLASS property can be correlated to
  a name or StartupWMClass key in a .desktop entry file, then
  the associated icon will be used.
- If the application has indicated an icon resource through WM_HINTS
  property, then the associated X window or pixmaps will be used.

Upstream Wine usually deals with this by placing a .desktop file with
StartupWMClass filled to match first string in WM_CLASS property
(which is the name of exe file being run).

Wine in Proton does not do it, but still puts "Wine" as second string,
therefore desktop environment can't differentiate between Wine in
Proton and Wine installed in OS.

By replacing "Wine" with "steam_app_<appid>" we force DE to fallback
to icon indicated by WM_HINTS (ico file embedded in exe file).
Steam can override this behaviour by installing properly crafted
.desktop entry file.  If SteamAppId environment variable is missing,
then generic "steam_proton" name is used instead.

### DIFF
--- a/dlls/winex11.drv/window.c
+++ b/dlls/winex11.drv/window.c
@@ -870,10 +870,13 @@ static void set_initial_wm_hints( Display *display, Window window )
     /* class hints */
     if ((class_hints = XAllocClassHint()))
     {
-        static char wine[] = "Wine";
+        static char steam_proton[] = "steam_proton";
+        char proton_app_class[128];
+        char *app_id = getenv("SteamAppId");
 
+        snprintf(proton_app_class, sizeof(proton_app_class), "steam_app_%s", app_id);
         class_hints->res_name = process_name;
-        class_hints->res_class = wine;
+        class_hints->res_class = app_id ? proton_app_class : steam_proton;
         XSetClassHint( display, window, class_hints );
         XFree( class_hints );
     }


dreamer:
This commit fixes ValveSoftware/Proton#1502 (at least "icon part" of that bug - "title part" comes from .desktop file detected by Gnome, but it doesn't really matter as Gnome will remove this UI in 3.32 anyway).

This fix affects mostly users of desktop environments with window managers derived from Mutter. Behaviour in KDE is not directly affected. Other DEs usually disregard WM_CLASS and use icon straight from WM_HINTS (sometimes in a buggy manner, but that's another story).

By using string `steam_app_<appid>` for WM_CLASS, Steam client can push integration with Linux DEs even further by crafting specific .desktop entry - this might be important because icon embedded in exe file might be completely different than one indicated by application developer for usage in Steam (consider e.g. DOSBOX games, old games missing hi-res icons, redesigned icons, etc). I will describe details in a separate bug report dedicated to that issue.

dreamer:
Now some screenshots to compare effect:

### Gnome

Behaviour in Proton 4.2-3 - notice broken Wine icon next to "Activities" and in bottom right corner of alt-tab UI. Icon being broken is either Fedora or Gnome bug (inability to interpret svg containing multiple images)

2019-04-25-210715_1366x768_scrot

Proton 4.2-3 with patch applied (screenshot was made with earlier version of patch, that's why there is "steam-proton" next to "Activities"). Window is now using Vault Boy icon, because that's the icon embedded in .exe file (which is ok). Ideally, icon should be [this one instead](https://steamcdn-a.akamaihd.net/steamcommunity/public/images/apps/38400/d5729a08544e28eaed2f48d7bae0f89a84cd1df0.ico), as [indicated by publisher](https://steamdb.info/app/38400/info/) (see "clienticon" link).

2019-04-25-211729_1366x768_scrot

### Cinnamon

Cinnamon uses it's own icon theme - generic Wine icon is this weird red thingy here, I guess:

2019-04-25-202350_1366x768_scrot

After patch:

2019-04-25-203158_1366x768_scrot

### KDE

Only one screenshot - icons looks the same before and after the patch

2019-04-25-200300_1366x768_scrot

KDE behaviour is somewhat complicated - depending on UI element, sometimes it respects the first string in WM_CLASS and sometimes does not. It never respects second string in WM_CLASS (as far as I can tell). In certain situations it leads to 2 different icons being visible at the same time - here's an example (notice icon in alt-tab UI and icon in the taskbar are different):

2019-04-25-200426_1366x768_scrot

aeikum:
Cool, thanks for doing all this research. My first thought was that we could just not set a class hint at all, or, always set it to "steam_proton" without dynamically creating the string. I think this would solve the icon problem, but I guess some WMs may use the class for other things, like saving the window position or something. So there is an advantage to setting it to "steam_app_1234". Right?

If so, then I think this idea is OK. Steam appids can be quite long (especially external apps), so I'd tweak the length of proton_app_class to be 512 bytes or something. I'd probably also have used the strn* functions instead of printf, but that's not a big deal.

dreamer:
> My first thought was that we could just not set a class hint at all, or, always set it to "steam_proton" without dynamically creating the string.

That was my first approach (hence "steam-proton" string next to Activities menu in Gnome screenshot), but according to [ICCCM protocol](https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.2.5), WM_CLASS is "mandatory". Gnome seems to ignore this part and displays icons when WM_CLASS is missing, but I don't know how other DEs behave. Tested it with game Osmos (29180) and it is literally the only one piece of software I could find, that does not set WM_CLASS prop.

According to the spec I linked above, the second string identifies the application and the first one identifies window within the application (not these words are used, but that's the gist).

> So there is an advantage to setting it to "steam_app_1234". Right?

Yes! :) DEs can use it for various things, but the application I had in mind was installing high-res icons in line with [XDG desktop entry spec](https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html). Usually, it's used to create "desktop shortcuts" and such, but it can be used as well to define application icon only - it's respected by at least KDE, Gnome and Cinnamon.

Example: after clicking "Create desktop shortcut", Steam puts .desktop file on Desktop, unpacks images from ico file, converts them to png and installs in `~/.local/share/icons`. After that Steam *could* create file `~/.local/share/applications/steam-fallout.desktop` with following content:

    [Desktop Entry]
    Type=Application
    Name=Fallout
    Icon=steam_icon_38400
    StartupWMClass=steam_app_38400
    NoDisplay=true

This would cause DE to pick better icons - often much better than the ones embedded in the .exe file. But it's a job of the Steam client - Proton can't do that, because only Steam has access to high-resolution icons uploaded by publisher and Steam is responsible for cleaning up during uninstall.

> I'd tweak the length of proton_app_class to be 512 bytes or something.

The longest App ID on Steam is 7 characters right now, but yeah - I'll bump it up at least to 128, should cover any reasonable App ID passed by a user.

aeikum:
If you add an external game, it gets a larger appid. I don't know how long they can get, but some random one I just installed is 10 digits long. Anyway 128 sounds fine.

aeikum:
This was merged as aef0f1638ce13db03a4d3c0fe53972f44443c854 in 4.2-4. Thanks again!

PR #48, opened 2019-05-17
Update to FAudio 19.06

flibitijibibo:
This uses the new functions introduced by FAudio 19.06, F3DAudioInitialize8 and FAudio_CommitOperationSet. The former's pretty boring, it just produces a real return code if a game gives us bogus initialize data, but the latter is basically required if games are going to keep working with FAudio 19.06 onward.

Operation sets can be extremely sensitive and affect a large portion of the XAudio2 spec, so the now-implemented operations will never execute and cause a memory leak and missing audio cues if the old CommitChanges is used. Similarly, we cannot actually do anything in FAudio_CommitChanges since the function cannot make assumptions about which sets to execute, so the function has been left as a stub with a single error message telling developers to update to CommitOperationSet.

Games affected include Warframe, DOOM 3 BFG, RAGE, Wolfenstein TNO, Wolfenstein TOB, XACT titles not using FACT (?), and potentially other titles using XAudio2 directly (I don't think Wwise uses operation sets).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1642,6 +1642,10 @@ then
     WINE_PACKAGE_FLAGS(FAUDIO,[faudio],[-lFAudio],,,
         [AC_CHECK_HEADERS([FAudio.h],
             [WINE_CHECK_SONAME(FAudio,FAudioCreate,,,[$FAUDIO_LIBS],[[libFAudio*]])])])
+    WINE_CHECK_LIB_FUNCS(\
+        FAudio_CommitOperationSet \
+        F3DAudioInitialize8,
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Patch has been made safe to use with older FAudio versions, so is now safe to merge at any time.

flibitijibibo:
Wine upstream version of this commit: https://source.winehq.org/git/wine.git/commit/d341efeff071532801fdc224062eb9e60dc5aea0

flibitijibibo:
This has been merged as of 4.2-6.

PR #49, opened 2019-05-18
libwine: Add WINEMODPATH variable

pchome:
New path priorities:
* WINEMODPATH
* builtin WINE libs
* WINEDLLPATH

A rough copy/paste hack to use custom `*.dll.so` libraries in the first place.

#### Use cases:
* DXVK (winelib), `WINEMODPATH="/path/to/dxvk/64:/path/to/dxvk/32" wine game.exe`
  No scripts, no overrides, etc.
* Built-in WINE libraries patched for a specific game.

I using this method for a while, for both vanilla WINE and custom Proton builds, and it work flawlessly for my needs. If you plan to return winelib DXVK -- something like this should be useful.

_Feel free to reject this PR. It's a PoC, "beautiful" patch rather than "beautiful" code, just to show what was changed._

### DIFF
--- a/libs/wine/loader.c
+++ b/libs/wine/loader.c
@@ -105,8 +105,8 @@ extern const char *get_dlldir( const char **default_dlldir );
 /* build the dll load path from the WINEDLLPATH variable */
 static void build_dll_path(void)
 {
-    int len, count = 0;
-    char *p, *path = getenv( "WINEDLLPATH" );
+    int len, count = 0, modcount = 0;
+    char *p, *path = getenv( "WINEDLLPATH" ), *modpath = getenv( "WINEMODPATH" );
     const char *dlldir = get_dlldir( &default_dlldir );
 
     if (path)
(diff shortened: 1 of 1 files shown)


aeikum:
I could be wrong, but I thought WINEDLLPATH already had priority over searching the install path. If that's the case, isn't it sufficient to prepend your new paths to WINEDLLPATH?

pchome:
Last time I checked this  WINEDLLPATH works this way only for "new/external" libs, e.g. adding path to `nvapi` from staging WINE for vanilla WINE.
For "native/internal" libs this don't work, e.g. adding path to DXVK winelib. In this case WINE's builtin `d3d11.dll.so` have a priority.

But I'll retest this again, maybe something changed in recent WINE versions.

aeikum:
If you're testing dxvk stuff in Proton, do be careful. We mess with WINEDLLOVERRIDES and WINEDLLPATH in the proton script to set that stuff up, so it may be the proton script that is interfering with your config attempts, not Wine.

pchome:
Yeah, I know. I mess even more with proton script and [protonfixes](https://github.com/simons-public/protonfixes), changing even more stuff. :)
I mostly know how both tools work.
 
I going to check on vanilla Wine, on something simple, to get short and reproducible result.

pchome:
So, I have wine-staging-4.8 and wine-proton-4.2 installed. Both patched to support WINEMODPATH.
Current `wine` is wine-staging-4.8.

```sh
cd /tmp && mkdir pfx
WINEPREFIX=/tmp/pfx wine init
WINEPREFIX=/tmp/pfx wine winecfg.exe
```
`winecfg` for **staging 4.8** shown
```sh
mkdir mod
cp /usr/lib64/wine-proton-9999/wine/winecfg.exe.so mod/
WINEPREFIX=/tmp/pfx WINEDLLPATH="/tmp/mod" wine winecfg.exe
```
`winecfg` for **staging 4.8** shown

```sh
WINEPREFIX=/tmp/pfx WINEMODPATH="/tmp/mod" wine winecfg.exe
```
`winecfg` for **proton 4.2** shown

pchome:
For `nv*.dll.so`(nvapi) WINEDLLPATH good enough, as usual.
Same approach, but this time `wine` is `wine-proton`,  `/tmp/pfx2` and /tmp/mod2/{32,64}.

`WINEDLLPATH="/tmp/mod2/32:/tmp/mod2/64"` works fine for `wine-proton`, because it don't contain any `nv*.dll.so`.

EDIT: well, "works fine" is not fully true, seems `*.dll.so` not fully compatible between versions (4.8 -> 4.2). But works in general.

PR #50, opened 2019-05-19
Fix rotating camera in AC: Unity and AC: Rogue

AlexeyProkhin:
Some Ubisoft games use DIPROP_GUIDANDPATH property to detect if a dinput device is also an xinput device. This patch fixes the gamepad support for me for following games:
1. AC Unity: fixes rotating camera;
2. AC Rogue: ditto (but the game requires [user32-ShowWindow patch](https://github.com/wine-staging/wine-staging/blob/master/patches/user32-ShowWindow/0001-user32-ShowWindow-should-not-send-message-when-windo.patch) from wine staging);
3. Far Cry 5: fixes vehicle steering  (but you probably would not be able to test it, because it requires to bypass EAC anticheat by replacing its dll with a version from other game with activated wine support; the games works in offline mode after that);
4. I don't have a steam copy of AC: Black Flag, but from reading ProtonDB reports, the patch is also relevant for the game as well.

### DIFF
--- a/dlls/dinput/joystick_sdl.c
+++ b/dlls/dinput/joystick_sdl.c
@@ -51,6 +51,9 @@
 #include "device_private.h"
 #include "joystick_private.h"
 
+#include "initguid.h"
+#include "devguid.h"
+
 #ifdef HAVE_SDL2_SDL_H
 
 WINE_DEFAULT_DEBUG_CHANNEL(dinput);
@@ -70,6 +73,7 @@ struct SDLDev {
     WORD vendor_id;
     WORD product_id;
     CHAR *name;
+    BOOL is_xbox_gamepad;
 
     BOOL has_ff, is_joystick;
     int autocenter;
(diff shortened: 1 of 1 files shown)


kisak-valve:
Hello @AlexeyProkhin, this patch should also be sent upstream if it hasn't been already.

AlexeyProkhin:
@kisak-valve Last I checked, the SDL dinput backend was not upstreamed. So the patch could not be also.

aeikum:
@AlexeyProkhin Thanks, nice work! We will test this and look at including it in an upcoming Proton version. I think this can go upstream by applying a similar change to the other backends that are in Wine. I can do this work to get it upstream if you do not want to.

AlexeyProkhin:
@aeikum The only thing I am not sure about is how to check if the device is an xinput gamepad for other backends.

AlexeyProkhin:
@aeikum I did a quick look over other backends yesterday and I think I would not have much trouble adapting the patch for them. I will try to do it soon and send the patches upstream.

AlexeyProkhin:
v2: set guidClass to GUID_DEVCLASS_HIDCLASS

aeikum:
This was merged as 9ea2c9c69bdcb970c6828b5375ca0796b96f862a in 4.2-5. Unfortunately your guidClass fixes came too late for this release. Please feel free to open a new PR to fix it.

PR #52, opened 2019-05-23
kernel32: Allow double quote style escape in argv.

redmcg:
Two quotes together (within outer quotes) represents a single
quote (with the first quote acting as an escape character)

This patch (accepted to Wine 4.4) allows Elite Dangerous to run without the need to install vcrun2015 (via winetricks). With this patch applied, only dotnet40 is required. And with the release of many WPF components in .NET Core - I'm hoping wine-mono will soon be able to remove even this requirement (and make ED Platinum).

Wine-Bug: https://bugs.winehq.org/show_bug.cgi?id=46721
Signed-off-by: Brendan McGrath <brendan@redmandi.com>
Signed-off-by: Alexandre Julliard <julliard@winehq.org>

### DIFF
--- a/dlls/kernel32/process.c
+++ b/dlls/kernel32/process.c
@@ -1506,9 +1506,13 @@ static char **build_argv( const UNICODE_STRING *cmdlineW, int reserved )
             /* '\', count them */
             bcount++;
         } else if ((*s=='"') && ((bcount & 1)==0)) {
-            /* unescaped '"' */
-            in_quotes=!in_quotes;
-            bcount=0;
+            if (in_quotes && s[1] == '"') {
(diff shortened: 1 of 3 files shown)


aeikum:
Thanks, will apply in an upcoming Proton release.

aeikum:
This slipped through the cracks for 4.2-6. I've picked it for the next release. Apologies for the delay.

redmcg:
No problem. Whilst dotnet40 is required, it's rather trivial to install vcrun2015.

redmcg:
I can see this patch made it in to 4.2-8; so I will close this PR. Thanks Andrew.

PR #55, opened 2019-06-13
fix issue for resident evil 4 and AC 4 black flag

jomarocas:
https://bugs.winehq.org/show_bug.cgi?id=39731#c8

https://bugs.winehq.org/show_bug.cgi?id=46336#c8


aeikum:
Assassin's Creed 4: Black Flag should be working now in Proton 5.13. I'm not sure about RE4.

jomarocas:
RE4 working always with version up to 5.0.1, close the issue

PR #62, opened 2019-08-18
Fix missing loader lock release in GetModuleHandleExW()

imaami:
[The return statement on line 482 in `dlls/kernel32/module.c`](https://github.com/ValveSoftware/wine/blob/aebdb22bb57128a5f38cd07d3952f1eb62e44817/dlls/kernel32/module.c#L482) may be executed without the loader lock set earlier never being released.


ryao:
This fixes a regression introduced by 79f64262e2a.

ryao:
@kisak-valve This patch fixes a regression that had been introduced by a developer at CodeWeavers in a proton specific patch. It might be a good idea to add the cw label to this PR.

aeikum:
@imaami Good catch, thanks! Will merge.

aeikum:
This was merged as 57f18487fb073df1adf3fcdb0726175d530aa737. Thanks!

PR #69, opened 2019-11-13
Implement integer scaling in fshack

misyltoad:
This adds the env var PROTON_FSHACK_INTEGER_SCALING to do nearest neighbour upscaling for fshack.

May help out some users playing pixel art titles at lower resolutions for UI scale or other reasons.

### DIFF
--- a/dlls/winevulkan/vulkan.c
+++ b/dlls/winevulkan/vulkan.c
@@ -1267,7 +1267,7 @@ VkResult WINAPI wine_vkGetPhysicalDeviceSurfaceCapabilitiesKHR(VkPhysicalDevice 
         capabilities->maxImageCount = 32;
 
     if(vk_funcs->query_fs_hack &&
-            vk_funcs->query_fs_hack(NULL, &user_res, NULL)){
+            vk_funcs->query_fs_hack(NULL, &user_res, NULL, NULL)){
         capabilities->currentExtent = user_res;
         capabilities->minImageExtent = user_res;
         capabilities->maxImageExtent = user_res;
@@ -1715,7 +1715,7 @@ VkResult WINAPI wine_vkCreateSwapchainKHR(VkDevice device, const VkSwapchainCrea
         our_createinfo.oldSwapchain = ((struct VkSwapchainKHR_T *)(UINT_PTR)our_createinfo.oldSwapchain)->swapchain;
 
     if(vk_funcs->query_fs_hack &&
-            vk_funcs->query_fs_hack(&object->real_extent, &user_sz, &object->blit_dst) &&
+            vk_funcs->query_fs_hack(&object->real_extent, &user_sz, &object->blit_dst, &object->fs_hack_filter) &&
             our_createinfo.imageExtent.width == user_sz.width &&
             our_createinfo.imageExtent.height == user_sz.height)
     {
@@ -1946,8 +1946,8 @@ static VkResult init_blit_images(VkDevice device, struct VkSwapchainKHR_T *swapc
     uint32_t blit_memory_type = -1, i;
 
     samplerInfo.sType = VK_STRUCTURE_TYPE_SAMPLER_CREATE_INFO;
-    samplerInfo.magFilter = VK_FILTER_LINEAR;
-    samplerInfo.minFilter = VK_FILTER_LINEAR;
+    samplerInfo.magFilter = swapchain->fs_hack_filter;
+    samplerInfo.minFilter = swapchain->fs_hack_filter;
     samplerInfo.addressModeU = VK_SAMPLER_ADDRESS_MODE_CLAMP_TO_BORDER;
     samplerInfo.addressModeV = VK_SAMPLER_ADDRESS_MODE_CLAMP_TO_BORDER;
     samplerInfo.addressModeW = VK_SAMPLER_ADDRESS_MODE_CLAMP_TO_BORDER;
@@ -2494,7 +2494,7 @@ static VkResult record_graphics_cmd(VkDevice device, struct VkSwapchainKHR_T *sw
     device->funcs.p_vkCmdBlitImage(hack->cmd,
             hack->user_image, VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
             hack->swapchain_image, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
-            1, &blitregion, VK_FILTER_LINEAR /* CUBIC_IMG? */);
+            1, &blitregion, swapchain->fs_hack_filter);
 
     /* transition user image from TRANSFER_SRC_OPTIMAL to GENERAL */
     barriers[0].sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER;
--- a/dlls/winevulkan/vulkan_private.h
+++ b/dlls/winevulkan/vulkan_private.h
@@ -165,6 +165,7 @@ struct VkSwapchainKHR_T
     VkDeviceMemory user_image_memory, blit_image_memory;
     uint32_t n_images;
     struct fs_hack_image *fs_hack_images; /* struct fs_hack_image[n_images] */
+    VkFilter fs_hack_filter;
     VkSampler sampler;
     VkDescriptorPool descriptor_pool;
     VkDescriptorSetLayout descriptor_set_layout;
--- a/dlls/winex11.drv/opengl.c
+++ b/dlls/winex11.drv/opengl.c
@@ -211,6 +211,7 @@ struct wgl_context
     struct gl_drawable *new_drawables[2];
     BOOL refresh_drawables;
     BOOL fs_hack;
+    BOOL fs_hack_integer;
     GLuint fs_hack_fbo, fs_hack_resolve_fbo;
     GLuint fs_hack_color_texture, fs_hack_ds_texture;
     GLuint fs_hack_color_renderbuffer, fs_hack_color_resolve_renderbuffer, fs_hack_ds_renderbuffer;
@@ -2091,6 +2092,7 @@ static void fs_hack_setup_context( struct wgl_context *ctx, struct gl_drawable *
 
         ctx->setup_for = p;
         gl->has_scissor_indexed = has_extension(glExtensions, "GL_ARB_viewport_array");
+        ctx->fs_hack_integer = fs_hack_is_integer();
         gl->fs_hack_context_set_up = TRUE;
     }
     else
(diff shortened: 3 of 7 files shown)


misyltoad:
Maybe I should rename this as it's not *true* integer scaling.

Could do some more work to ensure we upscale to the biggest multiple we can fit and black border around the rest of it though -- which would make it accurate.

aeikum:
This is a great idea. I probably won't have time to look at it for a few days or a week or two (sorting through DXGI-related regressions at the moment...). We are also making some fshack-related changes due to Zhiyi's work on multiple monitor support, so there may be some conflicts there. Anyway, cool stuff, will take a look as soon as I can.

aeikum:
> Maybe I should rename this as it's not _true_ integer scaling.
> 
> Could do some more work to ensure we upscale to the biggest multiple we can fit and black border around the rest of it though -- which would make it accurate.

I think that'd be a neat improvement, yeah.

misyltoad:
Is it worth me waiting for that other work to land first before I do that?

These changes were very basic and I don't mind re-making them.

aeikum:
That stuff won't land until the next rebase. If you want to do it now that should be fine, we'll port it over during the rebase.

misyltoad:
I have force pushed my implementation of true integer scaling.

aeikum:
Cool! Do you have a couple suggested games to try it out with?

misyltoad:
Force pushed some OpenGL optimizations for it

misyltoad:
Currently I have been testing with Half-Life 2

``PROTON_FSHACK_INTEGER_SCALING=1 PROTON_USE_D9VK=1 %command% -w 1024 -h 768 -fullscreen``

(1024x768 does not divide nicely into 1920x1080 or 3840x2160)

misyltoad:
What should happen:

IMG_20191120_190100

:tada:

zzhiyi:
This change doesn't seem too big. Let's pull it in before my multiple monitor support lands. I will do a rebase for it.

aeikum:
Filtered:

dx_fs

Integer scaling:

dx_integer

I like it!

One outstanding problem is how/when to turn this on. Obviously we can have user overrides for it, but does it make sense to have some attempt at auto-enabling it? Maybe something like if either dimension is less than half of the real dimension; or if `(desired X * desired Y) < (real X * real Y) / 4` or something. Have other projects encountered this? Do any classic console emulators have "smart" selection between filtering and pixel doubling?

misyltoad:
We wouldn't want to do this automatically... I think an option in Steam would be nice per-gane, with maybe a default in the appinfo.

aeikum:
You might be right, but I think it's at least worth thinking about. In what situations would a user want this enabled? One obvious application is older pixel games (although it seems most such releases on Steam already do their own scaling rather than mode switching), but we can't filter by that programmatically. I'd argue the unfiltered Deus Ex screenshot above looks better than the linear filtering, although it does have the downside of being both letter- and pillar-boxed.

I played around with the options a bit and less than half in *both* dimensions seems like a reasonable heuristic to me. I think it's also going to be a fairly rare case (even 800x600 is greater than half of my monitor's height), although high DPI monitors may be changing that.

    BOOL fs_hack_is_integer(void)
    {
        static int is_int = -1;
        if(is_int < 0)
        {
            const char *e = getenv("WINE_FULLSCREEN_SCALING");
            if(e && !strcmp(e, "pixel"))
                is_int = 1;
            else if(e && !strcmp(e, "smooth"))
                is_int = 0;
            else
                return (dd_modes[currentMode].width <= dd_modes[realMode].width / 2) &&
                    (dd_modes[currentMode].height <= dd_modes[realMode].height / 2);
        }
        return is_int;
    }

I dunno. Any thoughts?

pchome:
@aeikum 
> I'd argue the unfiltered Deus Ex screenshot above looks better than the linear filtering, although it does have the downside of being both letter- and pillar-boxed.

An option to force borderless-window mode would be great. Like `-popupwindow` command line argument in some Unity engine games ("-popupwindow (Windows only) | The window will be created as a a pop-up window (without a frame)").

Look and feel will depend on game itself and on user's preference. Pixel-art should look great, but user may also want to use scaling when e.g. on the couch, with big screen in other side of the room. So my vote for an option.

Also, I'm not sure about this patch, but I tested so-called "RETRO" mode from [libstrangle](https://gitlab.com/torkel104/libstrangle/blob/master/src/vulkan.c#L127) with Wasteland 2: DC, and it looked good in both scaling modes for me.

misyltoad:
@aeikum The problem with that is you get games that wanna use a mode very close to your display res that are pixel art. I dont think this is a good enough heuristic... Some people may dislike that we're upscaling this way by default in some titles where it doesn't really fit.

aeikum:
Yeah, that's a good point. I guess setting it manually (either the user, or us via app defaults) is the way to go.

aeikum:
I merged this with a couple trivial changes. The most significant is this change to your environment variable name:

    BOOL fs_hack_is_integer(void)
    {
        static int is_int = -1;
        if(is_int < 0)
        {
            const char *e = getenv("WINE_FULLSCREEN_INTEGER_SCALING");
            is_int = e && strcmp(e, "0");
        }
        return is_int;
    }

In general, I like to avoid using "Proton" from within Wine, as these commits may get applied to other projects (wine-staging, etc).

misyltoad:
You could make another function to get the value and do `static int is_int = fs_hack_is_integer_value();` and then return that to avoid the if statement.

(Assuming I haven't used a C++-ism there)

aeikum:
Good idea, I think that should work fine.

aeikum:
Or not. Was worth a shot.

    ../../../../proton/wine/dlls/winex11.drv/settings.c: In function 'fs_hack_is_integer':
    ../../../../proton/wine/dlls/winex11.drv/settings.c:285:25: error: initializer element is not constant
         static int is_int = check_integer_scaling_env();
                             ^
    make: *** [Makefile:575: settings.o] Error 1

misyltoad:
Yeah I thought it might be a C++ism...

Wine moving to C++ it is then :)))) 🐸🐸🐸🐸🐸

aeikum:
Heh. [Did you know `g++` intentionally breaks `dlclose`?](https://github.com/ValveSoftware/Proton/commit/bd8efaa9a63310382a92763fd44c9f8cfbabada7) That one was fun to debug. I wonder what other headaches that way lie ;)

pchome:
BTW, isn't env vars have "three states"?
* not set
* empty (`VAR=`)
* value (`VAR=0`)

defined / initialized ...

Sorry for the noise if I'm wrong.

kakra:
> I dunno. Any thoughts?

Yes, I have some: I'm using a UHD screen but need to scale a lot of games to QHD, and a few to FHD. FHD would be half resolution of UHD, so your heuristic may already kick in.

Let's assume something different: Either, modern pixel art games do their own scaling (or you could just run them at native or near-native resolution without performance problems), or they are very old and max out at some legacy resolutions. So the heuristic should look for resolutions below or equal to 1024x768, or for 4:3 resolution below 1/4th of the native pixel density. In pseudo-code:

```
if (res <= '1024x768' || (aspect <= 4/3 && running_megapixels <= native_megapixels/4)) { ... }
```

What do you think? Looking at the pixel count, a factor of 4 counts for half resolution as you already suggested, only that we also expect the aspect to be 4:3 (or less, so more quadratic than wide). Maybe we can tune the aspect ratio parameter a bit more. But I think legacy pixel-art games are either very low resolution or use 4:3 aspect ratio and somewhat low resolutions.

Still, I'm having some bad taste about this heuristic: It incorporates a value that is potentially only growing over time... The native screen resolution... We should strictly avoid such variables. Maybe better compare against static FHD pixel count, no matter what the future brings, e.g. `... running_megapixels <= FHD_megapixels`?

**UPDATE:** Let's compare to 8/5 instead of 4/3 so we would catch legacy resolutions of 320x200, too.

Marat-Tanalin:
## Heuristics idea: enable IS based on pixel density

One of reliable-enough heuristic approaches would probably be to enable integer scaling automatically if the native/logical ratio is lower than or equal to OS-level _pixel density_ (DPI, PPI).

For example, if the user has OS-level zoom set to 200% (2.0), it’s probably safe to assume that logical pixels are small enough to be almost indistinguishable at that ratio. This is the case e.g. when playing antialiased 3D games at FHD on a 24-27″ 4K monitor with 200% OS-level zoom.

In future, with a 8K monitor of the same size and OS-level zoom of 400% (4.0), integer scaling could be automatically enabled at any logical resolution resulting in a lower ratio such as 300% or 200%. So at OS-level zoom of 400%, we would get integer scaling in Proton automatically enabled for FHD (400%, logical pixels are small enough), QHD (300%, even smaller pixels), and 4K (200%, even more smaller pixels).

The behavior could be controlled via an option with the following list of possible values:

* Auto _[integer scaling is enabled heuristically according to pixel density and maybe something else]_
* Enabled _[integer scaling is enabled always, no heuristics magic]_
* Disabled _[regular blurry interpolation is used]_

The “Auto” value could probably be used by default.

aeikum:
This was merged as 157a4b8c0dada5020b73549f2eea4aea6838b1cc and is available in 4.11-10 with `WINE_FULLSCREEN_INTEGER_SCALING=1`. We can move the heuristic discussion to a new Issue or something, although I haven't yet seen a clear case where every user would prefer to use this mode. Maybe we should find one and base our discussion on that.

soredake:
@Joshua-Ashton is this working correctly?
with integer:
integer
without integer:
non_integer

misyltoad:
Yes.

misyltoad:
If you can't fit it in an integer scale factor, you get black borders around it.

Marat-Tanalin:
@soredake

`1080 / 600 = 1.8`, so the largest integer is `1`, and the rest `0.8` is filled with black, which is equivalent to regular centered mode.

For 800×600, the best-fit native heights are 1200 (2x, perfect fit) and 4320 (8K, 7x, just 3% of height lost).

Less optimal are 1440 (2x) and 4K (3x) with 17% of height lost in both cases.

1080 is only enough for 640×480 (2x, 11% of height lost) and lower resolutions.

Marat-Tanalin:
@Joshua-Ashton
Is the feature already available in the Proton version built into stable Steam?

kisak-valve:
Hello @Marat-Tanalin, yes, see https://github.com/ValveSoftware/wine/pull/69#issuecomment-565526767.

soredake:
@Joshua-Ashton @Marat-Tanalin thanks

switched to 640x480:
Screenshot_20200228_194049

soredake:
Same happens in dead space, but only after alt-tab, and cursor is doubled:
Screenshot_20200228_201307

pchome:
@Joshua-Ashton 
I guess you could try to emulate 21:9 with Wine virtual desktop [1720x720](https://en.wikipedia.org/wiki/Ultrawide_formats#Ultrawide_Cinema) resolution on regular monitor. Then test both integer and regular scaling mode with 720 and 360.

I can confirm, this is how (see above comment) integer scaling work on 21:9 monitor. Also, sometimes scaling is wrong in FS-hack mode, but I can't remember where and how.

PR #70, opened 2019-11-29
Fix some Force Feedback related bugs

berarma:
- Scale offset ~and phase~ in periodic effects.
- Set gain correctly
- Change calculation of effect direction

More info here: ValveSoftware/Proton#3194

I couldn't test it since I have never built Proton and I don't have the time now. Please, someone that can test it report on the results. Thanks!

### DIFF
--- a/dlls/dinput/effect_sdl.c
+++ b/dlls/dinput/effect_sdl.c
@@ -289,7 +289,7 @@ static HRESULT WINAPI effect_GetParameters(IDirectInputEffect *iface,
             DIPERIODIC *tsp = effect->lpvTypeSpecificParams;
 
             tsp->dwMagnitude = MulDiv(This->effect.periodic.magnitude, 10000, 32767);
-            tsp->lOffset = This->effect.periodic.offset;
+            tsp->lOffset = SCALE(LONG, 20000, -10000, This->effect.periodic.offset, 0xffff, -32767);
             tsp->dwPhase = This->effect.periodic.phase;
             tsp->dwPeriod = This->effect.periodic.period * 1000;
         }
--- a/dlls/dinput/effect_sdl.c
+++ b/dlls/dinput/effect_sdl.c
@@ -427,8 +427,6 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
 
     TRACE("%p %p 0x%x\n", This, effect, flags);
 
-    dump_DIEFFECT(effect, &This->guid, flags);
-
     if (IsEqualGUID(&This->guid, &GUID_Sine))
         This->effect.type = SDL_HAPTIC_SINE;
     else if (IsEqualGUID(&This->guid, &GUID_Triangle))
@@ -469,6 +467,8 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
         This->first_axis_is_x = effect->rgdwAxes[0] == DIJOFS_X;
     }
 
+    dump_DIEFFECT(effect, &This->guid, flags);
+
     if (flags & DIEP_DIRECTION)
     {
         if (effect->cAxes == 1)
@@ -479,7 +479,6 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
                 if (flags & DIEP_AXES)
                 {
                     SET_BASE_EFFECT_FIELD(This->effect, direction.dir[0], effect->rglDirection[0]);
-                    SET_BASE_EFFECT_FIELD(This->effect, direction.dir[1], effect->rglDirection[1]);
                 }
             } else {
                 /* one-axis effects must use cartesian coords */
@@ -567,7 +566,8 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
             tsp = effect->lpvTypeSpecificParams;
 
             This->effect.periodic.magnitude = MulDiv(tsp->dwMagnitude, 32767, 10000);
-            This->effect.periodic.offset = tsp->lOffset;
+            This->effect.periodic.offset = SCALE(Sint16, 0xffff, -32767, tsp->lOffset, 20000,
+                    -10000);
             This->effect.periodic.phase = tsp->dwPhase;
             if (tsp->dwPeriod <= 1000)
                 This->effect.periodic.period = 1;
--- a/dlls/dinput/joystick.c
+++ b/dlls/dinput/joystick.c
@@ -183,6 +183,13 @@ void dump_DIEFFECT(LPCDIEFFECT eff, REFGUID guid, DWORD dwFlags)
     TRACE("  - dwTriggerButton: %d\n", eff->dwTriggerButton);
     TRACE("  - dwTriggerRepeatInterval: %d\n", eff->dwTriggerRepeatInterval);
     TRACE("  - rglDirection: %p\n", eff->rglDirection);
+    if (dwFlags & DIEP_DIRECTION) {
+        TRACE("    %d", eff->rglDirection[0]);
+        if (eff->cAxes > 1 && eff->dwFlags & DIEFF_CARTESIAN) {
+            TRACE(" %d", eff->rglDirection[1]);
+        }
+        TRACE("\n");
+    }
     TRACE("  - cbTypeSpecificParams: %d\n", eff->cbTypeSpecificParams);
     TRACE("  - lpvTypeSpecificParams: %p\n", eff->lpvTypeSpecificParams);
 
--- a/dlls/dinput/joystick.c
+++ b/dlls/dinput/joystick.c
@@ -184,10 +184,9 @@ void dump_DIEFFECT(LPCDIEFFECT eff, REFGUID guid, DWORD dwFlags)
     TRACE("  - dwTriggerRepeatInterval: %d\n", eff->dwTriggerRepeatInterval);
     TRACE("  - rglDirection: %p\n", eff->rglDirection);
     if (dwFlags & DIEP_DIRECTION) {
-        TRACE("    %d", eff->rglDirection[0]);
-        if (eff->cAxes > 1 && eff->dwFlags & DIEFF_CARTESIAN) {
-            TRACE(" %d", eff->rglDirection[1]);
-        }
+        TRACE("    ");
+        for (i = 0; i < eff->cAxes; ++i)
+            TRACE("%d ", eff->rglDirection[i]);
         TRACE("\n");
     }
     TRACE("  - cbTypeSpecificParams: %d\n", eff->cbTypeSpecificParams);
--- a/dlls/dinput/joystick_sdl.c
+++ b/dlls/dinput/joystick_sdl.c
@@ -1295,11 +1295,13 @@ static BOOL _SetProperty(JoystickImpl *This, const GUID *prop, const DIPROPHEADE
         case (DWORD_PTR)DIPROP_FFGAIN:
         {
             LPCDIPROPDWORD pd = (LPCDIPROPDWORD)header;
-            int sdl_gain = MulDiv(This->sdldev->gain, 100, 10000);
+            int sdl_gain;
 
             TRACE("DIPROP_FFGAIN(%d)\n", pd->dwData);
 
             This->sdldev->gain = pd->dwData;
+
+            sdl_gain = MulDiv(This->sdldev->gain, 100, 10000);
 
             rc = SDL_HapticSetGain(This->sdldev->sdl_haptic, sdl_gain);
             if (rc != 0)


leillo1975:
This problem can be tested on rFactor 1 (effects missing), rFactor 2 (weak FFB and effects missing) and Automobilista (effects missing). Is a Proton exclusive problem. With Wine everything seems to be ok.

You must install [New-Lg4ff](https://github.com/berarma/new-lg4ff) to get FFB on this games. With the kernel Logitech Wheels driver they don't have FFB support. Kernel driver only implements Constant Force, and this is not included in some games like the related ones.

dlls/dinput/joystick_sdl.c
```diff
@@ -1295,12 +1295,14 @@ static BOOL _SetProperty(JoystickImpl *This, const GUID *prop, const DIPROPHEADE
...
         {
             LPCDIPROPDWORD pd = (LPCDIPROPDWORD)header;
-            int sdl_gain = MulDiv(This->sdldev->gain, 100, 10000);
+            int sdl_gain;
 
             TRACE("DIPROP_FFGAIN(%d)\n", pd->dwData);
 
             This->sdldev->gain = pd->dwData;
 
+            sdl_gain = MulDiv(This->sdldev->gain, 100, 10000);
```
flukejones, line 1304: You could remove the `int sdl_gain;` above and initialize it here?

```
int sdl_gain = MulDiv(This->sdldev->gain, 100, 10000);
```

flukejones:
> I couldn't test it since I have never built Proton and I don't have the time now. Please, someone that can test it report on the results. Thanks!

How do you know the changes are on the right track unless you test them yourself? I'll do a test build later and check.

berarma:
> How do you know the changes are on the right track unless you test them yourself? I'll do a test build later and check.

Yes, I know this is no good practice. I woudn't normally create a PR without testing the changes first but I thought it could be a better way to explain the bugs than just writing prose.

I'm fairly sure about the changes because I've been testing and debugging FF in games for days and I've been chasing these bugs for some time. I'm not so sure about the effect direction fix, but the current code it's pretty broken at least for wheels. I hope this change takes us a bit closer to fixing it.

I'll try to test them but it can take days or weeks until I get to do it.

dlls/dinput/joystick_sdl.c
berarma: Some C programming conventions recommend putting variable declarations at the start of blocks, that's why I've kept it there. I'm not sure what conventions Wine follows.

flukejones:
```
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c: In function 'effect_SetParameters':
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:482:37: error: 'SDL_HapticEffect {aka union SDL_HapticEffect}' has no member named 'direction'
                         This->effect.direction = 0x4000;
                                     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:484:37: error: 'SDL_HapticEffect {aka union SDL_HapticEffect}' has no member named 'direction'
                         This->effect.direction = 0xC000;
                                     ^
../../tools/winegcc/winegcc -o dispex.dll.fake -B../../tools/winebuild -b i686-w64-mingw32 --lib-suffix=.cross.a \
  -gsplit-dwarf -Wl,--wine-builtin -shared ../../../../proton/wine/dlls/dispex/dispex.spec \
  -mno-cygwin usrmarshal.cross.o dlldata.cross.o disp_ex_p.cross.o disp_ex_r.res \
  ../../dlls/uuid/libuuid.cross.a ../../dlls/oleaut32/liboleaut32.cross.a \
  ../../dlls/ole32/libole32.cross.a ../../dlls/rpcrt4/librpcrt4.cross.a \
  ../../dlls/advapi32/libadvapi32.cross.a  -Wl,--file-alignment,4096
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:486:37: error: 'SDL_HapticEffect {aka union SDL_HapticEffect}' has no member named 'direction'
                         This->effect.direction = 0;
                                     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:488:37: error: 'SDL_HapticEffect {aka union SDL_HapticEffect}' has no member named 'direction'
                         This->effect.direction = 0x8000;
                                     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:426:13: warning: unused variable 'retval' [-Wunused-variable]
     HRESULT retval = DI_OK;
             ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c: At top level:
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:532:5: error: expected identifier or '(' before 'if'
     if (flags & DIEP_DURATION)
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:548:5: error: expected identifier or '(' before 'if'
     if (flags & DIEP_STARTDELAY && effect->dwSize > sizeof(DIEFFECT_DX5))
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:553:5: error: expected identifier or '(' before 'if'
     if (flags & DIEP_TRIGGERBUTTON)
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:558:5: error: expected identifier or '(' before 'if'
     if (flags & DIEP_TRIGGERREPEATINTERVAL)
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:563:5: error: expected identifier or '(' before 'if'
     if (flags & DIEP_TYPESPECIFICPARAMS)
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:651:5: error: expected identifier or '(' before 'if'
     if (flags & DIEP_ENVELOPE)
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:669:5: error: expected identifier or '(' before 'if'
     if (flags & DIEP_GAIN)
     ^
In file included from ../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:36:0:
../../../../proton/wine/include/wine/debug.h:93:60: error: expected identifier or '(' before 'while'
     wine_dbg_log( __dbcl, __dbch, __FUNCTION__, args); } } while(0)
                                                            ^
../../../../proton/wine/include/wine/debug.h:90:8: note: in expansion of macro '__WINE_DBG_LOG'
        __WINE_DBG_LOG
        ^
../../../../proton/wine/include/wine/debug.h:472:36: note: in expansion of macro '__WINE_DPRINTF'
 #define WINE_TRACE                 __WINE_DPRINTF(_TRACE,__wine_dbch___default)
                                    ^
../../../../proton/wine/include/wine/debug.h:515:36: note: in expansion of macro 'WINE_TRACE'
 #define TRACE                      WINE_TRACE
                                    ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:670:9: note: in expansion of macro 'TRACE'
         TRACE("Effect gain requested but no effect gain functionality present.\n");
         ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:672:5: error: expected identifier or '(' before 'if'
     if (flags & DIEP_SAMPLEPERIOD)
     ^
In file included from ../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:36:0:
../../../../proton/wine/include/wine/debug.h:93:60: error: expected identifier or '(' before 'while'
     wine_dbg_log( __dbcl, __dbch, __FUNCTION__, args); } } while(0)
                                                            ^
../../../../proton/wine/include/wine/debug.h:90:8: note: in expansion of macro '__WINE_DBG_LOG'
        __WINE_DBG_LOG
        ^
../../../../proton/wine/include/wine/debug.h:472:36: note: in expansion of macro '__WINE_DPRINTF'
 #define WINE_TRACE                 __WINE_DPRINTF(_TRACE,__wine_dbch___default)
                                    ^
../../../../proton/wine/include/wine/debug.h:515:36: note: in expansion of macro 'WINE_TRACE'
 #define TRACE                      WINE_TRACE
                                    ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:673:9: note: in expansion of macro 'TRACE'
         TRACE("Sample period requested but no sample period functionality present.\n");
         ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:675:5: error: expected identifier or '(' before 'if'
     if (This->effect_id >= 0)
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:685:5: error: expected identifier or '(' before 'if'
     if (!(flags & DIEP_NODOWNLOAD))
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:687:5: error: expected identifier or '(' before 'if'
     if (retval != DI_OK)
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:690:5: error: expected identifier or '(' before 'if'
     if (flags & DIEP_START)
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:693:5: error: expected identifier or '(' before 'return'
     return DI_OK;
     ^
../../../../proton/wine/dlls/dinput8/../dinput/effect_sdl.c:694:1: error: expected identifier or '(' before '}' token
 }
```

dlls/dinput/effect_sdl.c
```diff
@@ -460,6 +460,8 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
                 DIEP_TRIGGERREPEATINTERVAL | DIEP_TYPESPECIFICPARAMS;
     }
 
+    dump_DIEFFECT(effect, &This->guid, flags);
```
flukejones: Should be unrequired as it is also on line 430

dlls/dinput/effect_sdl.c
```diff
@@ -474,12 +476,16 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
         if (effect->cAxes == 1)
         {
             if (effect->dwFlags & DIEFF_CARTESIAN)
-            {
```
flukejones: This `{` is required.

berarma:
I'll leave the the effect direction fix for another PR. Thanks for testing.

flukejones:
> I'll leave the the effect direction fix for another PR. Thanks for testing.

I'm testing some modification to what you submitted earlier. I only have Project Cars 2 as a non-working FFB game to test however. If it works I'll paste a git patch here.

berarma:
> I'm testing some modification to what you submitted earlier. I only have Project Cars 2 as a non-working FFB game to test however. If it works I'll paste a git patch here.

The problem with that game is that the effect direction is in the vertical axis so it's ignored by the wheel. I included a second `dump_DIEFFECT` call because the first one doesn't show information about the effect direction parameters, the flags need to be updated for the information to show. I wanted that information to try to understand what was happening.

flukejones:
This is what I did that sems to have fixed Project Cars 2 FFB for me (G920 wheel)

```
diff --git a/dlls/dinput/effect_sdl.c b/dlls/dinput/effect_sdl.c
index 8055785cc0..fcb89419aa 100644
--- a/dlls/dinput/effect_sdl.c
+++ b/dlls/dinput/effect_sdl.c
@@ -476,10 +476,23 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
             if (effect->dwFlags & DIEFF_CARTESIAN)
             {
                SET_BASE_EFFECT_FIELD(This->effect, direction.type, SDL_HAPTIC_CARTESIAN);
-                if (flags & DIEP_AXES)
-                {
-                    SET_BASE_EFFECT_FIELD(This->effect, direction.dir[0], effect->rglDirection[0]);
-                    SET_BASE_EFFECT_FIELD(This->effect, direction.dir[1], effect->rglDirection[1]);
+                if (flags & DIEP_AXES) {
+                    if (effect->rgdwAxes[0] == DIJOFS_X && effect->rglDirection[0] >= 0)
+                    {
+                        SET_BASE_EFFECT_FIELD(This->effect, direction.dir[0], 0x4000);
+                    }
+                    else if (effect->rgdwAxes[0] == DIJOFS_X && effect->rglDirection[0] < 0)
+                    {
+                        SET_BASE_EFFECT_FIELD(This->effect, direction.dir[0], 0xC000);
+                    }
+                    else if (effect->rgdwAxes[0] == DIJOFS_Y && effect->rglDirection[0] >= 0)
+                    {
+                        SET_BASE_EFFECT_FIELD(This->effect, direction.dir[0], 0);
+                    }
+                    else if (effect->rgdwAxes[0] == DIJOFS_Y && effect->rglDirection[0] < 0)
+                    {
+                        SET_BASE_EFFECT_FIELD(This->effect, direction.dir[0], 0x8000);
+                    }
                 }
             } else {
                 /* one-axis effects must use cartesian coords */
@@ -567,8 +580,9 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
             tsp = effect->lpvTypeSpecificParams;
 
             This->effect.periodic.magnitude = MulDiv(tsp->dwMagnitude, 32767, 10000);
-            This->effect.periodic.offset = tsp->lOffset;
-            This->effect.periodic.phase = tsp->dwPhase;
+            This->effect.periodic.offset = SCALE(Sint16, 0xffff, -32767, tsp->lOffset, 20000,
+                    -10000);
+            This->effect.periodic.phase = SCALE(Uint16, 0xffff, 0, tsp->dwPhase, 35999, 0);
             if (tsp->dwPeriod <= 1000)
                 This->effect.periodic.period = 1;
             else
diff --git a/dlls/dinput/joystick_sdl.c b/dlls/dinput/joystick_sdl.c
index 78e6f08391..80d8d559d9 100644
--- a/dlls/dinput/joystick_sdl.c
+++ b/dlls/dinput/joystick_sdl.c
@@ -1295,12 +1295,14 @@ static BOOL _SetProperty(JoystickImpl *This, const GUID *prop, const DIPROPHEADE
         case (DWORD_PTR)DIPROP_FFGAIN:
         {
             LPCDIPROPDWORD pd = (LPCDIPROPDWORD)header;
-            int sdl_gain = MulDiv(This->sdldev->gain, 100, 10000);
+            int sdl_gain;
 
             TRACE("DIPROP_FFGAIN(%d)\n", pd->dwData);
 
             This->sdldev->gain = pd->dwData;
 
+            sdl_gain = MulDiv(This->sdldev->gain, 100, 10000);
+
             rc = SDL_HapticSetGain(This->sdldev->sdl_haptic, sdl_gain);
             if (rc != 0)
                 ERR("SDL_HapticSetGain (%i -> %i) failed: %s\n", pd->dwData, sdl_gain, SDL_GetError());
```

Built version [here](https://drive.google.com/open?id=1AGjv190ga37zlv08fIzjUdicKcKKSABf)

flukejones:
I will try out your other changes a bit later. Feel free to ping me to test anything now that I have builds and other bits going.

berarma:
Thanks. I've added your patch. I'll try your builds when I can. Some driving games have demos that can be used to check the fixes, like Race07 and rFactor.

berarma:
Sorry for the multiple rebases. I've done it again because I wanted to start clean. This time I've tested the changes and it works in Project Cars 2 Demo, Race07 Demo, rFactor and rFactor 2. This set of games showed all the issues I'm trying to fix.

Changes:
- The periodic effect phase didn't need to be scaled. SDL2 already does it.
- The direction didn't need to be converted to an angle like Wine does since SDL2 uses a cartesian direction too. But setting the second axis was causing problems it seems, so I removed the assignement to the second axis.

I've changed the debug code a bit so that it would give me information about axis and direction. I can remove these changes if needed but I thought they could help debug future issues.

The Proton build can be downloaded [here](https://wetransfer.com/downloads/7fc1c242658f63967e2738cd6339b50d20191201225620/cac1dea6038d66b54b17fded79f7dbe020191201225620/13322c)

berarma:
@flukejones, I tried your build and it gave me reversed FFB in rFactor2. I could finally find some time and built Proton myself to try to find the reason. I found out the way I was trying to translate the direction parameter was wrong for SDL2 and thus your work based on mine was also wrong.

Sorry for making a hurried PR, I didn't know when or if I could do it myself and I wanted something to come along my explanations in ValveSoftware/proton#3194. Please, report your results if you test the latest changes. Thanks for your help.

aeikum:
Wow! Great work. We have a G27 and a DFGT here in the office. I'll review your changes and have our QA test the games you've mentioned.

leillo1975:
I tested the @berarma Proton Build and now Automobilista, RACE 07, rFactor 1 and 2 works great. We feel more effects and rF2 has a strong FFB

berarma:
@aeikum, sadly, only Project Cars 2 will work correctly when using the Logitech FFB in-kernel module for the DFGT and G2X wheels because it only supports constant force. Some of the fixes are for the periodic effects used on some of these games so they need a new driver that is in the works at [new-lg4ff](https://github.com/berarma/new-lg4ff). G920 users don't have this issue since all the effects are directly supported by the firmware.

This module adds all the missing effects that are supported in the Windows driver from Logitech. It's already working very well for many users with zero crashes, so you're welcome to try it. @leillo1975 has tested these fixes with the new driver under Proton and Wine with the same results.

We hope to see the fixes soon available for everyone. Thanks!

aeikum:
Understood. We'll confirm that games at least continue to work as they do now with the current kernel module (i.e. no regressions). I'm not sure if we'll have time to confirm the status with the new module, but it sounds like you are already on top of that.

dlls/dinput/effect_sdl.c
```diff
@@ -479,7 +479,6 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
                 if (flags & DIEP_AXES)
                 {
                     SET_BASE_EFFECT_FIELD(This->effect, direction.dir[0], effect->rglDirection[0]);
-                    SET_BASE_EFFECT_FIELD(This->effect, direction.dir[1], effect->rglDirection[1]);
```
aeikum, line 482: If setting `direction.dir[1]` to a value caused a problem, perhaps we should zero it? Ditto for `direction.dir[2]`, and all the other codepaths here.

dlls/dinput/joystick.c
```diff
@@ -183,6 +183,13 @@ void dump_DIEFFECT(LPCDIEFFECT eff, REFGUID guid, DWORD dwFlags)
     TRACE("  - dwTriggerButton: %d\n", eff->dwTriggerButton);
     TRACE("  - dwTriggerRepeatInterval: %d\n", eff->dwTriggerRepeatInterval);
     TRACE("  - rglDirection: %p\n", eff->rglDirection);
+    if (dwFlags & DIEP_DIRECTION) {
+        TRACE("    %d", eff->rglDirection[0]);
+        if (eff->cAxes > 1 && eff->dwFlags & DIEFF_CARTESIAN) {
+            TRACE(" %d", eff->rglDirection[1]);
```
aeikum: I think we should print all of the rglDirection values, similar to what we do with rgdwAxes above.

aeikum (commented):
Mostly looks good, some review comments below. Let me know what you think.

dlls/dinput/effect_sdl.c
berarma: The struct we're filling is initialized to zeroes in the function `sdl_create_effect`. And it's forbidden to change the number of axes after it's set so it seems safe to assume the non-used values will stay zeroes. The [DIEFFECT docs](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ee416616(v=vs.85)) say this: `The number of axes for an effect cannot be changed once it has been set.`

dlls/dinput/joystick.c
berarma: Yes, that would be better.

berarma:
I've noticed that the `GetParameters` also needed a range correction in the `offset` parameter of the periodic effect.

aeikum:
@berarma Thanks, everything looks great. I've merged your changes and split them up into individual patches. You can review my changes at https://github.com/aeikum/wine/commits/proton_4.11-ff-fixes (also contains some unrelated upcoming commits).

If you would like to contribute changes to upstream Wine, I think the changes to logging rglDirection and using dwFlags correctly in `effect_linuxinput.c` could go upstream. Or I can send it on your behalf.

berarma:
Thanks @aeikum. I'm pleased to contribute. Please, send them on my behalf because it will easily get lost in my queue of things to do. I haven't ever submitted any code to Wine.

Looking forward to seeing these fixes in the official Proton. You're doing a really great job.

aeikum:
@berarma So I'm looking at this SCALE macro, and it's totally wrong, isn't it?

    [aeikum@aeikum ~]$ cat test.c
    #include <stdio.h>

    #define SCALE(type, target_range, target_min, value, source_range, source_min) \
        (type)((((target_range)*(value + source_min))/source_range)-target_min)

    #define SCALE_FIXED(type, target_range, target_min, value, source_range, source_min) \
        (type)((((target_range)*(value - source_min))/source_range)+target_min)

    int main(int argc, char ** argv)
    {
        printf("orig 15/[10-20) to [0-100): %d\n",
                SCALE(int, 100, 0, 15, 10, 10));
        printf("fixed 15/[10-20) to [0-100): %d\n",
                SCALE_FIXED(int, 100, 0, 15, 10, 10));
        return 0;
    }
    [aeikum@aeikum ~]$ gcc -o test test.c
    [aeikum@aeikum ~]$ ./test
    orig 15/[10-20) to [0-100): 250
    fixed 15/[10-20) to [0-100): 50
    [aeikum@aeikum ~]$

aeikum:
I think this implementation fixes it

    -#define SCALE(type, target_range, target_min, value, source_range, source_min) \
    -    (type)((((target_range)*(value + source_min))/source_range)-target_min)
    +static int SCALE(int target_range, int target_min, int value, int source_range, int source_min)
    +{
    +    return MulDiv(target_range, value - source_min, source_range) + target_min;
    +}

berarma:
@aeikum Yes, you're right! But how then is the FFB working pretty well? I've done some more checking and it seems this macro works well for some use cases, coincidentally the ones in this file. Even if it's working well, it's totally counterintuitive and can lead to errors in the future.

I think both of your proposed solutions are much better.

berarma:
Yesterday I couldn't test the proposed changes. I'll try to test them later.

aeikum:
@berarma I think it was working fine for cases where both target_min and source_min are zero. For anything else, it was broken, though it could still return results in a valid range depending on the inputs.

It would be nice to confirm that things are still working well for you with SCALE fixed, just to be sure we weren't somehow relying on the broken behavior.

aeikum:
I've updated my branch to contain all the fixes I'm aware of. https://github.com/aeikum/wine/commits/proton_4.11-ff-fixes

berarma:
Building... Why a function instead of a macro? It's a one-liner and it's called several times inside a function.

aeikum:
Because macros are obnoxious and we use an optimizing compiler. It gets inlined.

    000000000002b670 <effect_GetParameters>:
       2b670:       55                      push   %rbp
       2b671:       48 89 e5                mov    %rsp,%rbp
       2b674:       41 55                   push   %r13
       2b676:       49 89 d5                mov    %rdx,%r13
       2b679:       41 54                   push   %r12
       2b67b:       45 89 c4                mov    %r8d,%r12d

       ...

       2bd84:       41 b8 ff 7f 00 00       mov    $0x7fff,%r8d
       2bd8a:       ba 10 27 00 00          mov    $0x2710,%edx
       2bd8f:       4d 8b 65 40             mov    0x40(%r13),%r12
       2bd93:       e8 10 cc fe ff          callq  189a8 <MulDiv>
       2bd98:       41 b8 ff ff 00 00       mov    $0xffff,%r8d
       2bd9e:       b9 20 4e 00 00          mov    $0x4e20,%ecx
       2bda3:       41 89 04 24             mov    %eax,(%r12)
       2bda7:       0f bf 53 4a             movswl 0x4a(%rbx),%edx
       2bdab:       81 c2 ff 7f 00 00       add    $0x7fff,%edx

berarma:
That's cool! Thanks for the teaching.

I've tried Race07 with your branch. It doesn't work well. The wheel is unexpectedly kicking hard from time to time.

I'm starting to think that the original macro was on point doing the right thing, although it isn't valid for any set of ranges. It maps any source range spanning negative and positive numbers to another range also spanning negative and positive numbers. Usually, [-10000, 10000] to [-32768, 32767] and the reverse. It works perfectly for those cases. There are some cases where `MulDiv` is used instead of `SCALE` where the macro would have failed. So I think whoever wrote the macro knew it would work only for some cases.

The new `SCALE` does the conversion from [-10000, 10000] to [-32767, 32768]. This produces an overflow on variables of type `Sint16`. I've confirmed this by narrowing the target range by one:

```
static int SCALE(int target_range, int target_min, int value, int source_range, int source_min)
{
    return MulDiv(target_range - 1, value - source_min, source_range) + target_min;
}
```

With this change we have a conversion from [-10000, 10000] to [-32767, 32767]. It's off by one but the overflow is avoided. I've tried with Race07 and rFactor2 and it works well so it's a confirmation that the problem is basically the overflow. I dont' say this is the fix, just a confirmation.

We could put a warning on the original macro to avoid someone mistakenly swapping the existing `MulDev` with the macro, as anyone might think they do the same. Or we should fix the new function to avoid the overflow, and ideally use the full range like the macro does.

aeikum:
Is the problem that the most minimum Sint16 is actually -32768 and not -32767? So instead of decreasing the range by one, lower the min by one (or better yet, use `SDL_MIN_SINT16`). E.g.:

    -            This->effect.periodic.offset = SCALE(0xffff, -32767, tsp->lOffset, 20000, -10000);
    +            This->effect.periodic.offset = SCALE(0xffff, -32768 /* or SDL_MIN_SINT16 */, tsp->lOffset, 20000, -10000);

aeikum:
Similarly, the range on the 10k values is 20,001, not 20,000: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ee416604(v=vs.85)

Sigh

berarma:
> Is the problem that the most minimum Sint16 is actually -32768 and not -32767?

No, the problem is that the maximum value is 32768 (SDL_MAX_SINT16 +1).

> Similarly, the range on the 10k values is 20,001, not 20,000: https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ee416604(v=vs.85)

Yes, the macro is misleading in every way but it gets the calculation right for the cases here.

aeikum:
> > Is the problem that the most minimum Sint16 is actually -32768 and not -32767?
> 
> No, the problem is that the maximum value is 32768 (SDL_MAX_SINT16 +1).

What I mean is, the range [-10000,+10000] should map to [-3276**8**,+32767] (i.e. the range of Sint16). The previous macro usage specified -3276**7** as the minimum of this range. This means +10000 would map to +3276**8** which overflows, as you said.  So if we decrease the target minimum to -3276**8** we will end up with a maximum value of +3276**7**, which doesn't overflow Sint16.

Edit: Also, the too-small 10k range means +10,000 exceeds the maximum value for the source range [-10000,9999], which could exacerbate this even further.

aeikum:
I've updated my branch again to fix the scaling ranges, see aeikum/wine@a2fc06832ab5fba04569c35fb2ee2e2a964b76a6.  Can you double-check it and try it out?

berarma:
> I've updated my branch again to fix the scaling ranges, see aeikum/wine@a2fc068. Can you double-check it and try it out?

OK.

I was trying to rewrite the macro in a way that made more sense, just for the record:
```
#define SCALE_AROUND_ZERO(value, source_max, target_max) \
        ((((target_max * 2 + 1)*(value - source_max))/(source_max * 2)) + target_max)
```

EDIT: (this versions works in all cases)
```
#define SCALE_INT16_AROUND_ZERO(value, source_max) \
        (((0xffff *(value - source_max))/(source_max * 2)) + 0x7fff)
```

berarma:
> What I mean is, the range [-10000,+10000] should map to [-32768,+32767] (i.e. the range of Sint16). The previous macro usage specified -32767 as the minimum of this range. This means +10000 would map to +32768 which overflows, as you said. So if we decrease the target minimum to -32768 we will end up with a maximum value of +32767, which doesn't overflow Sint16.

Yes and no. The previous macro did it right.

> Edit: Also, the too-small 10k range means +10,000 exceeds the maximum value for the source range [-10000,9999], which could exacerbate this even further.

The source range is [-10,000, 10,000], so the range size is really 21,001.

EDIT: Sorry, some mistakes.

aeikum:
Thought about it some more and what we actually care about is the max value after adjusting the range to have 0 as its minimum. So 20,000 is actually correct. Will update my branch accordingly.

berarma:
> Thought about it some more and what we actually care about is the max value after adjusting the range to have 0 as its minimum. So 20,000 is actually correct. Will update my branch accordingly.

Think that a generic function won't take care of the asymmetrical case that is [-32768:32767] . The original macro did. `SCALE_INT16_AROUND_ZERO` is the same macro after narrowing down its use cases to the one it was written for.

aeikum:
OK yeah, I see what you mean. I think we can simplify a bit further, since we only have four possible conversions: centered to and from SDL, and zeroed to and from SDL. (Except for the deadband, which we can just handle specially). I hard-coded some macros for each case. I tested each macro to verify that zero always converts to zero, and that no valid input value will exceed the target range.

One possible downside is when going from SDL to DI (decreasing resolution), we can lose very small non-zero values. For example, 1 in SDL will convert to 0 in DI. I don't know if this is a problem in practice.

Anyway, take a look.  aeikum/wine@dd4539c0b6901901baed556f2603072846591e03.

berarma:
Yes, this is like the original macro but split in its 4 possible uses. It looks good.

There's one issue though. The `left_sat` and `right_sat` fields are Uint16, so the range should be [0, 65535].

Also, I think we can fill the whole range [-32768, 32767] without much effort. It's not as important as zero being zero, but it's an easy fix.

I've enclosed all macros in parentheses to avoid problems when used inside more complex expressions.

I've attached below a patch with the changes. I've tested them and it seems to work well.

> One possible downside is when going from SDL to DI (decreasing resolution), we can lose very small non-zero values. For example, 1 in SDL will convert to 0 in DI. I don't know if this is a problem in practice.

I think this problem already existed. The only way to return the exact same values that were previously sent would be saving the original values but I don't think it's worth it. The differences will be so small that the values will be practically the same.

The patch:
```
diff --git a/dlls/dinput/effect_sdl.c b/dlls/dinput/effect_sdl.c
index ae56dc8f00..c682e7d7bb 100644
--- a/dlls/dinput/effect_sdl.c
+++ b/dlls/dinput/effect_sdl.c
@@ -169,21 +169,21 @@ static HRESULT WINAPI effect_GetEffectGuid(IDirectInputEffect *iface, GUID *out)
         value = (target.custom.field); \
     }
 
-/* map [-32768, 0, 32767] -> [-9999, 0, 10000] */
+/* map [-32768, 0, 32767] -> [-10000, 0, 10000] */
 #define CENTERED_SDL_TO_DI(val) \
-    (20000 * (val + 0x7fff)) / 0xffff - 9999
+    ((20000 * (val - 0x7fff)) / 0xffff + 10000)
 
-/* map [-10000, 0, 10000] -> [-32767, 0, 32767] */
+/* map [-10000, 0, 10000] -> [-32768, 0, 32767] */
 #define CENTERED_DI_TO_SDL(val) \
-    (0xfffe * (val + 10000)) / 20000 - 0x7fff
+    ((0xffff * (val - 10000)) / 20000 + 0x7fff)
 
-/* map [0, 32767] -> [0, 10000] */
-#define ZEROED_SDL_TO_DI(val) \
-    (10000 * val) / 0x7fff
+/* map [0, sdl_max] -> [0, 10000] */
+#define ZEROED_SDL_TO_DI(val, sdl_max) \
+    ((10000 * val) / sdl_max)
 
-/* map [0, 10000] -> [0, 32767] */
-#define ZEROED_DI_TO_SDL(val) \
-    (0x7fff * val) / 10000
+/* map [0, 10000] -> [0, sdl_max] */
+#define ZEROED_DI_TO_SDL(val, sdl_max) \
+    ((sdl_max * val) / 10000)
 
 static HRESULT WINAPI effect_GetParameters(IDirectInputEffect *iface,
         DIEFFECT *effect, DWORD flags)
@@ -301,7 +301,7 @@ static HRESULT WINAPI effect_GetParameters(IDirectInputEffect *iface,
         {
             DIPERIODIC *tsp = effect->lpvTypeSpecificParams;
 
-            tsp->dwMagnitude = ZEROED_SDL_TO_DI(This->effect.periodic.magnitude);
+            tsp->dwMagnitude = ZEROED_SDL_TO_DI(This->effect.periodic.magnitude, 0x7fff);
             tsp->lOffset = CENTERED_SDL_TO_DI(This->effect.periodic.offset);
             tsp->dwPhase = This->effect.periodic.phase;
             tsp->dwPeriod = This->effect.periodic.period * 1000;
@@ -330,8 +330,10 @@ static HRESULT WINAPI effect_GetParameters(IDirectInputEffect *iface,
                 tsp[i].lOffset = CENTERED_SDL_TO_DI(This->effect.condition.center[i]);
                 tsp[i].lPositiveCoefficient = CENTERED_SDL_TO_DI(This->effect.condition.right_coeff[i]);
                 tsp[i].lNegativeCoefficient = CENTERED_SDL_TO_DI(This->effect.condition.left_coeff[i]);
-                tsp[i].dwPositiveSaturation = ZEROED_SDL_TO_DI(This->effect.condition.right_sat[i]);
-                tsp[i].dwNegativeSaturation = ZEROED_SDL_TO_DI(This->effect.condition.left_sat[i]);
+                tsp[i].dwPositiveSaturation = ZEROED_SDL_TO_DI(This->effect.condition.right_sat[i],
+                        0xffff);
+                tsp[i].dwNegativeSaturation = ZEROED_SDL_TO_DI(This->effect.condition.left_sat[i],
+                        0xffff);
                 tsp[i].lDeadBand = This->effect.condition.deadband[i] * 10000 / SDL_MAX_UINT16;
             }
         }
@@ -578,7 +580,7 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
                 return DIERR_INVALIDPARAM;
             tsp = effect->lpvTypeSpecificParams;
 
-            This->effect.periodic.magnitude = ZEROED_DI_TO_SDL(tsp->dwMagnitude);
+            This->effect.periodic.magnitude = ZEROED_DI_TO_SDL(tsp->dwMagnitude, 0x7fff);
             This->effect.periodic.offset = CENTERED_DI_TO_SDL(tsp->lOffset);
             This->effect.periodic.phase = tsp->dwPhase;
             if (tsp->dwPeriod <= 1000)
@@ -625,8 +627,10 @@ static HRESULT WINAPI effect_SetParameters(IDirectInputEffect *iface,
 
             for (i = j = 0; i < 3; ++i)
             {
-                This->effect.condition.right_sat[i] = ZEROED_DI_TO_SDL(tsp[j].dwPositiveSaturation);
-                This->effect.condition.left_sat[i] = ZEROED_DI_TO_SDL(tsp[j].dwNegativeSaturation);
+                This->effect.condition.right_sat[i] = ZEROED_DI_TO_SDL(tsp[j].dwPositiveSaturation,
+                        0xffff);
+                This->effect.condition.left_sat[i] = ZEROED_DI_TO_SDL(tsp[j].dwNegativeSaturation,
+                        0xffff);
                 This->effect.condition.right_coeff[i] = CENTERED_DI_TO_SDL(tsp[j].lPositiveCoefficient);
                 This->effect.condition.left_coeff[i] = CENTERED_DI_TO_SDL(tsp[j].lNegativeCoefficient);
                 This->effect.condition.deadband[i] = tsp[j].lDeadBand * SDL_MAX_UINT16 / 10000;
```

aeikum:
That does fix `CENTERED_DI_TO_SDL` to use the whole range, but it causes `0` to become `1` in `CENTERED_SDL_TO_DI`.

Good catch on the `sat` variables. Will apply those changes.

Edit: This version seems to do it right across the full range:

    #define CENTERED_SDL_TO_DI(val) \
        (20000 * (val - 0x8000)) / 0xffff + 10000

Updated my branch yet again...

berarma:
> Edit: This version seems to do it right across the full range:

Yes. Well done! It seems I focused my checks on the other function.

This time I think we're nailing it. I'll test your branch ~when ready~ to make sure there's no regressions.

berarma:
I'm testing the new build. I'll report on the results. No problems so far.

Anyone who wants to help test it but can't build it themselves can download the new build [here](https://we.tl/t-dSpVVrNHt2).

leillo1975:
I played [GT Legends](https://github.com/ValveSoftware/Proton/issues/2728) [RACE 07](https://github.com/ValveSoftware/wine/pull/70#issuecomment-562513442), [rFactor 1](https://github.com/ValveSoftware/Proton/issues/3307), [rFactor 2](https://github.com/ValveSoftware/Proton/issues/245#issuecomment-550071540), [Automobilista](https://github.com/ValveSoftware/Proton/issues/246), [Project Cars 1](https://github.com/ValveSoftware/Proton/issues/244) and [Project Cars 2](https://github.com/ValveSoftware/Proton/issues/908) without problems and without ffbtools fixes. Everything is ok

madbad:
Tested on a DFGT on Fedora 30 on rFactor1.
Seems to be working fine: the only issue I've encountered (not sure what is to blame here if the game itself or something else) it's if I reduce the in-game ForceFeedback effect to less than 100% the whole forcefeedback disappear.

berarma:
> Tested on a DFGT on Fedora 30 on rFactor1.
> Seems to be working fine: the only issue I've encountered (not sure what is to blame here if the game itself or something else) it's if I reduce the in-game ForceFeedback effect to less than 100% the whole forcefeedback disappear.

Thanks @madbad for testing. I'm testing it on a G29 and it works even when setting the FFB to less than 100%. I've tried 90%. Actually this game has to be set to -100% since the FFB is reversed. A setting of -90% works too. What are your other FFB settings?

madbad:
Just to keep things updated, we tracked down my issue to a bug that was already resolved by Berarma Logitec Driver on a more recent commit of what I was using.
So everything looks good from my side now FF seems to work perfectly everywhere.

aeikum:
This series was merged in 4.11-10. Thanks very much for working on this!

leillo1975:
[Project Cars](https://github.com/ValveSoftware/Proton/issues/244) , [Project Cars 2](https://github.com/ValveSoftware/Proton/issues/908) and [Kartkraft](https://github.com/ValveSoftware/Proton/issues/2886) has again FFB with Proton 4.11-10. In this games using [new-lg4ff](https://github.com/berarma/new-lg4ff) now we have a great FFB:
-[Automobilista](https://github.com/ValveSoftware/Proton/issues/246)
-[rFactor](https://github.com/ValveSoftware/Proton/issues/3307)
-[rFactor 2](https://github.com/ValveSoftware/Proton/issues/245)
-[RACE 07](https://github.com/ValveSoftware/Proton/issues/1212)
-[GT Legends](https://github.com/ValveSoftware/Proton/issues/2728)

[Dirt Rally 2](https://github.com/ValveSoftware/Proton/issues/2366) still needs [ffbtools](https://github.com/berarma/ffbtools) with "--update-fix" 

[Wreckfest](https://github.com/ValveSoftware/Proton/issues/758) has the same problem that using 4.11-9,  If total degrees are a segment, the center is in 80% of this segment instead of 50%. With Proton 4.2 the game works great

PR #71, opened 2019-12-05
HACK: ntdll: Return a handle to kernel32 when being asked for mfc42.

### LINKED ISSUE #2624: Add MFC42.dll "loader hack", which helps multiple games.
# Feature Request

## I confirm:
- [x] that I haven't found another request for this feature.
- [x] that I have checked whether there are updates for my system available that
      contain this feature already.

## Description 

Add "loader hack" to pretend to load ```mfc42```:
1. During ```LoadLibrary```detect if app uses ```mfc42```.
1. Return ```kernel32``` handle when app requests ```mfc42```.

This solves the ```pidgen.dll``` issue, and should allow the games to run.
*Full credit goes to @aeikum for this proposal.*

## Games for which a stub would fix the issue:
#17 (https://github.com/ValveSoftware/Proton/issues/2624#issuecomment-562839599)

## Games which seem to need a full implementation of *MFC42.dll*:
#1593
#1805
#2623
#2728

## Related Wine bugreport:
https://bugs.winehq.org/show_bug.cgi?id=657

Sbte:
For some applications mfc42 is loaded, but never actually used. We can add Steam game IDs when we find more of such applications.

Closes ValveSoftware/Proton#2624

Note that there are more games mentioned in that issue, but I haven't verified that they are actually fixed by this hack, so I didn't include those in the patch.

### DIFF
--- a/dlls/ntdll/loader.c
+++ b/dlls/ntdll/loader.c
@@ -159,6 +159,18 @@ static inline void ascii_to_unicode( WCHAR *dst, const char *src, size_t len )
     while (len--) *dst++ = (unsigned char)*src++;
 }
 
+static WCHAR *strcasestrW( const WCHAR *str, const WCHAR *sub )
+{
+    while (*str)
+    {
+        const WCHAR *p1 = str, *p2 = sub;
+        while (*p1 && *p2 && tolowerW(*p1) == tolowerW(*p2)) { p1++; p2++; }
+        if (!*p2) return (WCHAR *)str;
+        str++;
+    }
+    return NULL;
+}
+
 #define RTL_UNLOAD_EVENT_TRACE_NUMBER 64
 
 typedef struct _RTL_UNLOAD_EVENT_TRACE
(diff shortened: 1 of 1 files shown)


Sbte:
Actually, this should probably be done only after the dll was not found, so a user can still always supply their own mfc42. I'll fix the patch later and also test it for the other games.

Sbte:
I updated the patch and confirmed that it only does something useful for AoE3. The other games mentioned in the Proton issue still crash because of missing mfc42 functionality. There are other older games that may benefit from this, but they are not available on Steam.

aeikum:
Looks good, will merge.

leillo1975:
GT Legends needs this dll:
https://github.com/ValveSoftware/Proton/issues/2728
it would be good not to install it separately. I hope you will find a solution

Sbte:
I tested GT Legends with the hack (as stated in ValveSoftware/Proton#2624), and it doesn't help. It needs an actual implementation.

aeikum:
This was merged as 304683886016ec940a9b331ce33fcbaf47e893c3.

PR #72, opened 2019-12-05
Update bcrypt, apply pending ECDSA patches, and add hack for older GNUTLS versions.

Guy1524:
This merge request should get the Master Chief Collection working OOTB on Proton 4.11.

In the last commit, we fall back to _gnutls_decode_ber_rs_raw, an internal function, if the newer, identical, documented function gnutls_decode_rs_value isn't present.  This internal function is present on the version of GNUTLS shipped by Ubuntu 18.04, however it is not yet present in GNUTLS 3.4.10, which is shipped in the Steam Runtime.

It is possible to commit [2409d03e](https://gitlab.com/gnutls/gnutls/commit/2409d03e10a0588361db3cc2ee8d0fab94c7fc33) to this version of the library though, which should make the problem go away.  However, I haven't found this to be necessary, as Proton seems to use the system version of GNUTLS over the Steam Runtime version.


aeikum:
Nice work, will merge.

TTimo:
Next Steam Runtime update will have the function (>= 20191210.1)

aeikum:
This was merged in 4.11-10. With the Steam Beta client, Halo:MCC should be playable for all users, even on older distros.

PR #73, opened 2019-12-18
Unset LD_PRELOAD before executing explorer.exe to disable gameoverlayrenderer.so.

### LINKED ISSUE #3316: 4.11-10 update today = horrendous mouse stutter/lag after about 20 minutes
since installing the new 4.10-11 I'm now getting horrendous mouse stutter/lag after about 20 minutes of gameplay in Skyrim Special Edition (not tried any other games yet.

Skyrim is using 4.11-10 by default.
Linux fedora30 5.3.15-300.fc31.x86_64 #1 SMP Thu Dec 5 15:04:01 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
nvidia GTX 980 with nvidia 440.36 drivers.
I do not get this behaviour at the desktop.

rbernon:
This should fix https://github.com/ValveSoftware/Proton/issues/3316 until gameoverlayrenderer.so is patched.


tannisroot:
Wow.

PR #76, opened 2020-02-06
(HACK) Forward certain environment variables to the system for steamclient.so.

Guy1524:
Pierre-Loup found out that EALink is setting the SteamAppId environment variable before setting up it's connection to steamclient.  The variable wasn't being picked up by the Linux steamclient.so library, so the behavior of GetAuthSessionTicket was incorrect.  This PR forwards the two environment variables set by EALink to the system through setenv, so that they are picked up by native steamclient library.

### DIFF
--- a/dlls/ntdll/env.c
+++ b/dlls/ntdll/env.c
@@ -36,6 +36,7 @@
 #include "winnt.h"
 
 WINE_DEFAULT_DEBUG_CHANNEL(environ);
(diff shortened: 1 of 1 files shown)


aeikum:
Awesome, will merge.

Guy1524:
Looks like the better solution is to sync the environment variables on the steamclient side.

PR #77, opened 2020-02-08
Add WINEUSER environment variable

qsniyg:
This is a modified mirror of [my patch to wine](https://source.winehq.org/patches/data/178254). This allows proton to be used on existing prefixes while sharing data from the home directory, without needing symlinks or copies (by using the `WINEUSER=...` environment variable).

On a sidenote, `GetUserNameExW` from secur32 and `GetNamedPipeHandleStateW` from kernel32 also call `wine_get_user_name`, which would return the real username instead of steamuser, so this patch fixes that as well :)

(Note: I haven't actually tested this specific patch, but I have tested the one I sent to wine)

### DIFF
--- a/dlls/advapi32/advapi.c
+++ b/dlls/advapi32/advapi.c
@@ -84,7 +84,7 @@ GetUserNameA( LPSTR lpszName, LPDWORD lpSize )
 BOOL WINAPI
 GetUserNameW( LPWSTR lpszName, LPDWORD lpSize )
 {
-    const char *name = "steamuser"/*wine_get_user_name()*/;
+    const char *name = wine_get_user_name();
     DWORD i, len = MultiByteToWideChar( CP_UNIXCP, 0, name, -1, NULL, 0 );
     LPWSTR backslash;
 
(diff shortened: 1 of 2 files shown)


aeikum:
What's the usecase you're trying to fix here? Proton isn't intended to be used on non-Proton prefixes. There might be a better way to do what you're trying to do.

qsniyg:
@aeikum It's useful for the cases I wrote in the patch. Most recently, I ended up with a case where a game I had used in a wine prefix was working better in proton than in wine, but since the username was changed, all the configuration was lost. In this specific instance, renaming the username's folder to steamuser caused issues because it had hardcoded paths to my username. Symlinks do work, but it's far from an elegant solution.

I'm curious though, is there a reason for patching "steamuser" only in `GetUserNameW`, rather than in `wine_get_user_name`?

aeikum:
This is a non-Steam game, then? Or, you wanted to preserve your game data from a win32 Steam wineprefix into a Proton prefix? How did you run Proton in this wineprefix?

kakra:
> There might be a better way to do what you're trying to do.

Symlinks...

I suggest that people put symlinks if they want to share profiles between prefixes or even within the same prefix. Tho, it's a different kettle of fish whether this may be a good idea

leg0las-devel:
I ran into a problem, when running new beta Eve Online launcher say me about "i'm running under administrator, please run under normal user". As i understood **steamuser** has admin role, and need little fix to possibility run exe files as normal user (btw normal wine or wine-staging use as username system (linux) username (sorry for tautology). In future lack of this opportunity will break launch Eve Online.

tried run on wine-proton 7.0.6 and 8.0.

PR #80, opened 2020-02-25
fshack: Create only one compute pipeline per swap chain

DadSchoorse:
There is no need to create one pipeline per swapchain image.

PR #81, opened 2020-02-25
winevulkan: Cleanup barriers for fs hack

misyltoad:
Removes needless GENERAL transitions for graphics presentation.
(we can be TRANSFER_SRC for vkCmdBlitImage, no need to disable DCC with GENERAL)
Access masks were also incorrect for this, this fixes those. (VK_ACCESS_TRANSFER_READ_BIT was missing)

Fixes some spec violations when presenting from compute queues:
Previously, we were transitioning to PRESENT_SRC too early and copies were happening to PRESENT_SRC which is illegal.
This patch will fix that by them being GENERAL (they are storage images anyway) at this point, then transitioning.
Also fixes misc. access flag issues.

### DIFF
--- a/dlls/winevulkan/make_vulkan
+++ b/dlls/winevulkan/make_vulkan
@@ -2512,7 +2512,7 @@ class VkGenerator(object):
         f.write("     * resolution; user_sz will contain the app's requested mode; and dst_blit\n")
         f.write("     * will contain the area to blit the user image to in real coordinates.\n")
         f.write("     * All parameters are optional. */\n")
-        f.write("    VkBool32 (*query_fs_hack)(VkExtent2D *real_sz, VkExtent2D *user_sz, VkRect2D *dst_blit);\n")
+        f.write("    VkBool32 (*query_fs_hack)(VkExtent2D *real_sz, VkExtent2D *user_sz, VkRect2D *dst_blit, VkFilter *filter);\n")
 
         f.write("};\n\n")
 
--- a/dlls/winevulkan/make_vulkan
+++ b/dlls/winevulkan/make_vulkan
@@ -171,7 +171,6 @@ FUNCTION_OVERRIDES = {
     "vkCmdExecuteCommands" : {"dispatch" : True, "driver" : False, "thunk" : False},
     "vkCreateCommandPool" : {"dispatch": True, "driver" : False, "thunk" : False},
     "vkDestroyCommandPool" : {"dispatch": True, "driver" : False, "thunk" : False},
-    "vkCmdPipelineBarrier" : {"dispatch" : True, "driver" : False, "thunk" : False},
     "vkDestroyDevice" : {"dispatch" : True, "driver" : False, "thunk" : False},
     "vkFreeCommandBuffers" : {"dispatch" : True, "driver" : False, "thunk" : False},
     "vkGetDeviceProcAddr" : {"dispatch" : False, "driver" : True, "thunk" : False},
--- a/dlls/winevulkan/vulkan.c
+++ b/dlls/winevulkan/vulkan.c
@@ -1510,7 +1510,6 @@ static VkResult create_descriptor_set(VkDevice device, struct VkSwapchainKHR_T *
 static void destroy_fs_hack_image(VkDevice device, struct VkSwapchainKHR_T *swapchain, struct fs_hack_image *hack)
 {
     device->funcs.p_vkDestroyPipeline(device->device, hack->pipeline, NULL);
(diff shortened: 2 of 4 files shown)


aeikum:
Thanks, will take a look soon. @Joshua-Ashton could you review #80 as well? I think you're more qualified to review Vulkan stuff than me ;)

DadSchoorse:
there is still one issue, the user image doesn't get created with `VK_IMAGE_USAGE_TRANSFER_SRC_BIT`, but we want to use it as a blit src.

misyltoad:
Yeah I am aware of that problem still existing. Ideally we'd want to hook into the swapchain creation and add that usage flag.

DadSchoorse:
we do that, but we need to add `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` https://github.com/ValveSoftware/wine/blob/013a9b234732a9fbbd77ca370a3d66cf587a9113/dlls/winevulkan/vulkan.c#L1593

misyltoad:
Oh, didn't know we already touched that. Alright yeah that should be straight forward to solve then :P

DadSchoorse:
Also, the swapchain needs to be created with only `USAGE_TRANSFER_DST` and with `USAGE_STORAGE_BIT`. ofc without `USAGE_STORAGE_BIT` if the surface doesn't support it. https://github.com/ValveSoftware/wine/blob/013a9b234732a9fbbd77ca370a3d66cf587a9113/dlls/winevulkan/vulkan.c#L1743

aeikum:
I've pulled this in for the next release. Happy to pull in more if you have further changes (please don't change the commits I've already pulled :) )

DadSchoorse:
It just occurred to me that render passes can change the image layout, too. So we can't be sure that the user image is in `TRANSFER_SRC`. Also it's not really possible to change the image layout on render pass creation because we don't know which image is used with it.

So my solution would be to stop changing the layout in the image barriers and letting the application transition it to `PRESENT`. That way we can be totally sure that it's in that image layout before we do the blit/compute shader. It's also not against spec to transition a non swapchain image to `PRESENT`. Of course that means we need to put one memory barrier before the blit to transition to `TRANSFER_SRC` and we need to transition the user image to `PRESENT` again in the end.
@Joshua-Ashton do you know another solution?

misyltoad:
Hmmmmmmm... Can you transition in this way via a render pass on its own *from* ``PRESENT``? I'll check the spec when I am home.

DadSchoorse:
I'm pretty sure you can.
And imo we should just stop changing the barriers, it's just too big of a hack. I created https://github.com/Joshua-Ashton/proton-wine/pull/1 to remove them, but I'm not sure if the whole function `wine_vkCmdPipelineBarrier` could be removed.

misyltoad:
Thanks, I merged that PR into my branch. Looks good.

Ideally it'd be cool if we can get rid of that whole function. (as per our Discord discussion)

I'm going to diff fshack and see what other stuff I can find later today, as well as hopefully make some changes to WineVulkan that I want to make for Proton specifically.

misyltoad:
@aeikum The last couple of commits should have a good perf boost for all Vulkan (inc. DXVK) titles on Proton. Would be cool to get that pulled in 👌

aeikum:
You bet, thanks!

misyltoad:
When I was reviewing them I also saw that there was a critical section there that was eliminated by doing this. We really don't want a critical section on every barrier 😛 

I have some other perf changes I want to make soon to winevulkan more specifically: I'm going to close this now as this fulfills everything I wanted from this.

aeikum:
Let's leave this open until it's shipped.

misyltoad:
Sure 🐸

DadSchoorse:
I have created https://github.com/Joshua-Ashton/proton-wine/pull/2 to generate the pipeline barrier function again.

misyltoad:
I merged that

aeikum:
@Joshua-Ashton @DadSchoorse I've pushed a branch to https://github.com/aeikum/wine/tree/vulkan-fshack that I think contains everything from this PR and #80 . Can you please verify it looks all good to you? Assuming it passes QA, that's what's heading into the next release.

misyltoad:
Hi @aeikum,

Looks good to me for the next release :+1:

I have some more winevulkan perf improvements but I would like to save those until the release after next and until I have done extensive testing on them (given they're pretty huge changes.)

I will PR those (winevulkan-opt) after this next release ships. Don't want to feature creepy break stuff with a million and one changes either.

aeikum:
This was merged in the series ending in a4f4dd8c6578bb153a06487d071cbd790a83aa3a. Thanks!

PR #82, opened 2020-02-27
WIP: winevulkan: implement Contrast Adaptive Sharpening scaling in fshack

DadSchoorse:
This is work in progress.
CAS scaling (or Fidelity Fx CAS like AMD wants us to call it) should be able to preserve more details when upscaling. Currently you can activate it with the `WINE_FULLSCREEN_CAS_SCALING` env var. It works via a compute shader.
todo:

- any api other than vulkan
- clarify if the shader binary should stay in vulkan.c or if it should get it's own header
- figure out a good way to set the sharpness (a float between 0.0 and 1.0) via a env var or something else
- deal with different input and output aspect ratios
- ~~verify that the shader is  working correctly, currently there is a little bit of introduced grain~~ :heavy_check_mark:
- make sure that it works with an srgb swapchain


DadSchoorse:
Also this needs #81 and the things mentioned in the comments there.

kakra:
Wow, you are implementing what I just thought about a few days ago: Why doesn't Proton fshack implement CAS? That's timing!

Maybe we could even get some sort of AI upscaling at some point in the future ;-)

kisak-valve:
Hello @DadSchoorse, at a glance, this looks like something that should go to upstream wine first, then request that it be brought over with a smaller patch that has the fshack changes, but we'll need to hear from a Proton dev on how they want to handle this. I think that it would be a good idea to split this into two commits, the first having everything that can go upstream, and the second with Proton-specific changes.

DadSchoorse:
@kisak-valve as far as a know, there is nothing here that could be upstreamed since it's all changes to fshack.

DadSchoorse:
Now that I fixed the grain issue the shader binary got even longer, so I would prefer a separate header for it.

DadSchoorse:
Here is a quick comparison:
CAS:
grafik
bi-linear:
grafik

On the performance site there are still some things that can be done to increase speed.

DadSchoorse:
I came to the conclusion that I will close this pull request.
It's just too big for fshack which is already kinda fragile in itself.
Also I think there wouldn't be that many people actually using it.
Performance is kinda meh, too.
While the extra details are nice and it's superior to bilinear upscaling with post process CAS (e.g. via vkBasalt), it's not worth putting it directly into winevulkan.

PR #84, opened 2020-03-11
winevulkan: use swapchain format for fshack image + view

DadSchoorse:
Currently, fshack uses `RGBA8_UNORM ` for the tmp image and the storage image view, which is against spec since the format has to match the image format (unless `VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT` is set). This also fixes potential future issues when a game creates a non BGR8 swapchain.


DadSchoorse:
Actually, with this shader we would need to activate the shaderStorageImageWriteWithoutFormat device feature. Which by itself isn't a problem since every gpu/driver supports that, the only issue being that we would need to activate it always (even if fshack isn't active). So I'm not sure if that the right thing to do since every swapchain on linux will be rgba8 currently.

DadSchoorse:
This is actually no longer valid since the shaderStorageImageWriteWithoutFormat changes in 1.2.195.

PR #89, opened 2020-04-04
 Patch backport: Audio fixes for TrackMania Nations Forever

qwertychouskie:
See https://github.com/ValveSoftware/Proton/issues/3728.  Note that I don't have a build environment currently set up for Wine, but as these patches are from released Devel/Staging versions they should work just fine. :)

### DIFF
--- a/dlls/dsound/buffer.c
+++ b/dlls/dsound/buffer.c
@@ -27,7 +27,6 @@
 #include "winbase.h"
 #include "winuser.h"
 #include "mmsystem.h"
-#include "winternl.h"
 #include "vfwmsgs.h"
 #include "wine/debug.h"
 #include "dsound.h"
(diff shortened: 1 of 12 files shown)


kisak-valve:
https://source.winehq.org/git/wine.git/commit/13087008437d80542123fc3087cbef85f636a50a 

https://source.winehq.org/git/wine.git/commit/e09190399904c6ae13ce256aaa2efbf9360ac8c9

https://github.com/wine-staging/wine-staging/blob/master/patches/winmm-Replace_OpenFile/0001-winmm-Use-_lopen-instead-of-OpenFile.patch

aeikum:
Thanks for the suggestion. I will merge the dsound commits, but I'm still working with Alistair to get the winmm fix upstreamed. The patch in wine-staging is incorrect.

qwertychouskie:
@aeikum Is there any particular reason why the existing patch can't be applied in the interim?  Right now, I'm running Steam in Wine Staging, but it would be nice to just use native Steam and Proton :)

aeikum:
@qwertychouskie Because it's wrong and could break something we don't know about. In any case, it's been fixed and upstreamed now. I'll add it to the queue to backport for the next release. Thanks for the suggestion!

qwertychouskie:
@aeikum No problem!  Should I close this PR?

kisak-valve:
Helllo @qwertychouskie, let's keep this pull request open to track the request in progress.

qwertychouskie:
Got it, thanks :)

aeikum:
CC ValveSoftware/Proton#3387.

qwertychouskie:
Looks like this is in the latest RC now, I'll close this. :tada:

PR #91, opened 2020-05-18
winex11.drv: Support child window rendering for Vulkan via XComposite

stevenlafl:
In relation to https://github.com/ValveSoftware/Proton/pull/3886 and https://github.com/ValveSoftware/Proton/issues/3681 the Workshop Tools have child windows (the preview panes, Hammer views, etc.) that do not display: `006b:fixme:vulkan: X11DRV_vkCreateWin32SurfaceKHR Application requires child window rendering , which is not implemented yet!`

Upstream issue: https://bugs.winehq.org/show_bug.cgi?id=45277

### DIFF
--- a/dlls/winex11.drv/vulkan.c
+++ b/dlls/winex11.drv/vulkan.c
@@ -33,6 +33,7 @@
 #include "wine/heap.h"
 #include "wine/library.h"
 #include "x11drv.h"
+#include "xcomposite.h"
 
 #define VK_NO_PROTOTYPES
 #define WINE_VK_HOST
@@ -55,6 +56,7 @@ static CRITICAL_SECTION_DEBUG critsect_debug =
 static CRITICAL_SECTION context_section = { &critsect_debug, -1, 0, 0, 0, 0 };
 
 static XContext vulkan_hwnd_context;
+static XContext vulkan_swapchain_surface_context;
 
 static struct vulkan_funcs vulkan_funcs;
 
@@ -64,6 +66,7 @@ struct wine_vk_surface
 {
     LONG ref;
     Window window;
+    HDC child_window_dc;
     VkSurfaceKHR surface; /* native surface */
 };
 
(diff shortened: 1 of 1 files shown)


aeikum:
If I'm reading that upstream bug report correctly, this patch still has problems, right?

stevenlafl:
Yeah, you might have seen my comment and the one about Z order (which does appear to happen precisely as one predicted, if that's why child windows laid in top of each other, such as with Hammer's Top/Side/etc views appear to flash back and forth). I wanted to keep this here for posterity with the other PR but serves to shine a little more light on it as this functionality is pretty core to getting that (in my opinion flagship) product working right with wine/Proton.

It does -work-, but I am seeing is that basically only one child window appears to render at a time (so looks a little like flashing), like with having multiple 3d views together that both update as an object is being manipulated (such as in a modeling program where looking at both Top and Side views at the same time in different child windows). It just needs some tweaks. 

I'm by no means an expert in Wine, so if anyone has a comment on whether those two particular findings are indeed correct, Workshop Tools is certainly a fantastic test case for someone who is.

stevenlafl:
Placing this here for reference, see _Accelerated Compositing_ https://blogs.igalia.com/carlosgc/2016/09/20/webkitgtk-2-14/

jqadev:
@aeikum is this PR related to child window rendering mentioned in the Proton Experimetal changelog https://github.com/ValveSoftware/Proton/wiki/Changelog#available-in-proton-experimental-as-of-20210401
?

kisak-valve:
Hello @jqadev, you can compare https://github.com/ValveSoftware/wine/commit/9145d36ae4e5d4bd407f3ec9cdc844cfe45219c3 from the experimental_6.3 branch to the merge request and assess any changes yourself.

jqadev:
Thank you @kisak-valve. I missed it on a phone.

rbernon:
We have this merged in some different form now, so I'm closing this MR.

The version in proton could benefit from improvements, and https://github.com/ValveSoftware/wine/pull/152 tries to make it better, but the code now has some interactions with other hacks that makes it difficult to change. It'd probably be better redone from scratch with all the requirements in mind, and upstreamed.

PR #95, opened 2020-07-21
Proton 5.0 next

hamzamanaseer:
10011101010
east
then 
easy


kisak-valve:
Hello @hamzamanaseer, it's unclear what this pull request would have helped, but Proton 5.0-10 will be released when the Proton devs are happy with how it looks.

PR #109, opened 2021-05-12
Update winevulkan spec for experimental 6.3

liam-middlebrook:
Cherry picked patches from upstream WINE and re-generated commits where generated files were touched.


aeikum:
These are in experimental-6.3-20210514.

liam-middlebrook:
Closing out this PR since the commits have been added to the desired branch.

ThanosSiopoudis:
**Testing correction + rev 8 (supersedes the Testing section above):**

- The run I first cited as verifying the GL fix was contaminated: the `MESA_GL_VERSION_OVERRIDE=4.5COMPAT` diagnostic was still active in the device's user profile (caught via the `GL version 4.5 is not supported on wow64` clamp lines; the override also made wined3d emit `glScissorArrayv` for 16 scissors against the driver's 1 viewport after a settings change — the exact reason it can't ship).
- **Re-verified clean (override confirmed absent):** Syberia renders from the UI on real GL 3.1 — no GLSL/FBO/DSA errors, pad virtual-mouse works.
- That clean run surfaced one more 4.5-ism in the same path: unconditional sampler-object state save/restore around the present blit → per-frame `GL_INVALID_ENUM in glGetIntegeri_v(GL_SAMPLER_BINDING)` flood (cosmetic — no sampler objects exist on GL 3.1, so there is no state to restore). Patch 0006 regenerated with the save/restore gated on `ARB_sampler_objects`; `PROTON_REV=8`. **HW verification of the gate pending** (titan.8 package).
- Touch-offset on aspect-scaled titles = separate banked issue; pad virtual-mouse confirmed fine.

PR #110, opened 2021-05-15
winex11.drv: Fix fshack filter.

DadSchoorse:
Fixes a regression in 021848ce0a44ff600316e5e5e40c65bf1b33ee86 because filter would be not initialized.


aeikum:
@DadSchoorse Thank you very much. I've squashed that fix down into the faulty commit and released it in experimental-6.3-20210517, which is live now.

PR #112, opened 2021-06-03
wineboot: Return TSC frequency in ~Mhz

misyltoad:
## wineboot: Return TSC frequency in ~Mhz

Some games such as Horizon Zero Dawn use this registry value to correlate values from rtdsc -> real time.

In my testing across a few devices, Windows always returns the tsc frequency in this entry, not the current/maximum frequency of the processor.

Returning the nominal/maximum cpu frequency here causes the game to run in slow motion as it may not match the tsc frequency of the processor.

Ideally we'd not have to measure this and the kernel would return tsc_khz to userspace, but this is a good enough stop-gap until https://lkml.org/lkml/2020/12/31/72 or something similar is merged.

Fixes: #4125 (Slow motion bug)

Signed-off-by: Joshua Ashton <joshua@froggi.es>

## wineboot: Make freq in read_tsc_frequency static

Avoids needing to calculate TSC frequency multiple times. Looks like this was originally intended also as the if (freq == 0) would always return true.

Signed-off-by: Joshua Ashton <joshua@froggi.es>


programs/wineboot/wineboot.c
```diff
@@ -245,7 +245,7 @@ static void initialize_xstate_features(struct _KUSER_SHARED_DATA *data)
 
 static UINT64 read_tsc_frequency(void)
 {
-    UINT64 freq = 0;
+    static UINT64 freq = 0;
```
rbernon: The helper was initially used only once because the frequency was set in USD page.

As I've added the HV page later (to make sure USD and QPC freq is fixed at 10MHz as some code may expect it), it's been called twice as the HV page may benefit from a better precision.

IMHO it would be better to compute the frequency once keeping as much precision as possible and use it to set the various locations where it's used, instead of having a static variable.

rbernon: Also it's only going to work with Proton Experimental, there's no TSC frequency computation in Proton 6.3 yet (and I'm not sure there's going to soon, as I'm not completely satisfied with the calibration). I'm looking for ways to get it from the kernel, but there's no really good way to do so.

kakra: What if this could be made overwritable via an environment variable? People that feel like overriding the auto-detected value with one specific to their hardware could do so then.

misyltoad: Well, we can easily bring that function over for 6.3.

But yeah, it's sad there's no way of getting it, there were some patches for it but nothing ever in mainline.
Maybe we could have something for it in our own and community kernels though.

misyltoad: Made it calculate once at the start of wineboot and pass it to the right places instead of static-ing it

misyltoad: @kakra Might be worth doing as a separate patch, this TSC stuff is already used elsewhere.

aeikum:
I believe some version of this is in `experimental-6.3-20210604`.

kakra:
@Joshua-Ashton What about a user-space helper that can gain superuser access by asking polkit for permissions, which in turn then returns the TSC freq in the best way it could (just emitting it to stdout)? Feral's gamemode does something similar for switching powersafe modes by using a polkit-helper. If it is not installed or missing permissions, wine could still fall back to some other algorithm. This could be distributed as a separate package anyone could install on demand. One probably needs to figure out how this integrates with bwrap but I think it could be a good solution without carrying custom kernel patches.

misyltoad:
@kakra I think this is overkill, the TSC frequency should just be exposed to userspace in a nice way.

mirh:
I think there are 464 too many commits in this pull request..

https://github.com/trailofbits/tsc_freq_khz
http://lkml.iu.edu/hypermail/linux/kernel/1005.3/01184.html
Also, it's not exactly clear if TSC exporting is going to be accepted.

Sonicadvance1:
I just noticed this.
if you're running a recent generation Intel CPU you can calculate the TSC frequency using CPUID now. (Rocket Lake+ currently)
CPUID leaf 15h, if EAX, EBX, and ECX isn't zero then you can calculate the TSC frequency using `ecx * ebx / eax`.
Sadly AMD platforms don't support this.

rbernon:
I think this has been merged for a while, closing.

PR #116, opened 2021-07-18
winevulkan: Add FSR for fshack.

DadSchoorse:
This is an implementation of AMD's FSR https://gpuopen.com/fidelityfx-superresolution/ in fshack, which allows users to chose it instead of linear/nearest filtering for upscaling. It can be enabled using `WINE_FULLSCREEN_FSR=1` and a lower in game resolution than the display's native one. `WINE_FULLSCREEN_FSR_STRENGTH` controls the amount of sharpening (not to be confused with amd's quality presets, those are about the input resolution), 0 is the maximum sharpness, higher values mean less sharpening. ~~The default is 5.~~ The default is now 2 because that's AMD's rough recommendation.

Caveats:
- Only works in vulkan games (dxvk and vkd3d-proton included).
- Some games upscale on their own, without using fshack, when you select a lower resolution in game. FSR won't work in those.
- Obviously applying FSR to the final image is not ideal, it should be applied before the UI is drawn and before some post processing effects like film grain.
- AMD also recommends a negative LOD bias, which is not possible in a generic way without problems.
- This is only the slow fp32 version, if accepted I might follow up with the fp16 version.
- The patchset is a bit of a mess. If the huge commit is a problem I can try to split it up a bit.
- An environment variable to add custom resolution modes would be nice to more directly control the quality. e.g. 2048x1152


DadSchoorse:
I added another commit for a negative mip bias env var, you can use `WINE_VULKAN_NEGATIVE_MIP_BIAS=N` for a bias of `N/100`. For 1080p -> 1440p upscaling AMD recommends a negative bias between 0.38 and 0.58, so something like `WINE_VULKAN_NEGATIVE_MIP_BIAS=45`. By default, this is only applied to samplers with anisotropic filtering enabled, you can use `WINE_VULKAN_BIAS_ALL_SAMPLERS=1` to force it for all samplers.

**Do note that forcing a mip bias can break game graphics!**

gabriele2000:
> Replying to https://github.com/ValveSoftware/wine/pull/116#issue-692078584

Will it be possible to choose the various presets through and env var?
I'm currently playing the witcher 3 from 720p to 1080p and the result is very interesting, still I sense that we could do a lot more

DadSchoorse:
@gabriele2000 I'm not sure what you mean with presets, you control the quality by selecting the in game resolution and you adjust sharpness/mip bias a bit with the environment variables if you want.

gabriele2000:
Presets like quality, balanced, performance... More sharpening ≠ more
quality

Il gio 22 lug 2021, 10:43 Georg Lehmann ***@***.***> ha
scritto:

DadSchoorse:
Like I said, the quality presets of AMD just mean different game resolutions that get upscaled to your display's native resolution.

gabriele2000:
From 720p to 1080p what is the preset?

Il gio 22 lug 2021, 12:24 Georg Lehmann ***@***.***> ha
scritto:

palaashatri:
might sound like a noob question, but how will we be able to choose the ingame render resolution that will be upscaled to our native monitor resolution by fsr? for example, some old games like hitman blood money for example, which do not have an ingame resolution slider, running through dxvk?

Kron4ek:
> From 720p to 1080p what is the preset?

Quality, according to the info from [this](https://gpuopen.com/fidelityfx-superresolution/) page.

have-a-boy:
> might sound like a noob question, but how will we be able to choose the ingame render resolution that will be upscaled to our native monitor resolution by fsr? for example, some old games like hitman blood money for example, which do not have an ingame resolution slider, running through dxvk?

Afaik, that should just be the output resolution you set in the games' display options. With the caveat that some games implement their own upscaling from whatever resolution you set, to the target resolution of your window/screen, which prevents upscaling after that point.

dlls/winevulkan/vulkan.c
```diff
@@ -912,6 +914,26 @@ VkResult WINAPI __wine_create_vk_device_with_callback(VkPhysicalDevice phys_dev,
     if (res != VK_SUCCESS)
         goto fail;
 
+    /* Enable shaderStorageImageWriteWithoutFormat for fshack
+     * XXX check if available
+     */
```
aeikum, line 919: Is it difficult to add this check?

dlls/winevulkan/vulkan.c
```diff
@@ -1997,84 +2019,871 @@ void main()
...
+    if (any(greaterThanEqual(gl_GlobalInvocationID.xy, c4.zw)))
+        return;
+
+    FsrEasuF(color, uvec2(gl_GlobalInvocationID.xy), c1, c2, c3, c4);
+
+    imageStore(outImage, ivec2(gl_GlobalInvocationID.xy), vec4(color, 1.0));
+}
+*/
+const uint32_t fsr_easu_comp_spv[] = {
+    0x07230203,0x00010000,0x0008000a,0x0000129e,0x00000000,0x00020011,0x00000001,0x00020011,
```
aeikum, line 2110: Weird indent here, and in some of the other shader bytecodes.

dlls/winevulkan/vulkan.c
```diff
@@ -2096,7 +2905,8 @@ static VkResult create_descriptor_set(VkDevice device, struct VkSwapchainKHR_T *
     descriptorAllocInfo.pSetLayouts = &swapchain->descriptor_set_layout;
 
     res = device->funcs.p_vkAllocateDescriptorSets(device->device, &descriptorAllocInfo, &hack->descriptor_set);
-    if(res != VK_SUCCESS){
+    if (res != VK_SUCCESS)
+    {
```
aeikum: Please don't add unnecessary diffs, they make rebases even harder.

dlls/winex11.drv/vulkan.c
```diff
@@ -826,7 +826,7 @@ static VkSurfaceKHR X11DRV_wine_get_native_surface(VkSurfaceKHR surface)
 }
 
 static VkBool32 X11DRV_query_fs_hack(VkSurfaceKHR surface, VkExtent2D *real_sz, VkExtent2D *user_sz,
-        VkRect2D *dst_blit, VkFilter *filter)
+        VkRect2D *dst_blit, VkFilter *filter, BOOL *fsr, float *sharpness)
```
aeikum, line 829: This isn't a blocker for your series, but I wonder if we can make this API suck less (e.g. use a struct instead of a bunch of arguments).

dlls/winex11.drv/fs.c
```diff
@@ -535,6 +535,28 @@ BOOL fs_hack_is_integer(void)
...
+    if (sharpness)
+    {
+        const char *e = getenv("WINE_FULLSCREEN_FSR_STRENGTH");
+        if (e)
+        {
+            sharpness_int = atoi(e);
+        }
+        *sharpness = (float) sharpness_int / 10.0f;
+    }
+    TRACE("is_fsr: %s, sharpness: %2.4f\n", is_fsr ? "TRUE" : "FALSE", sharpness ? *sharpness : 0.0f);
```
aeikum: This feels bike-sheddy (sorry) but I think we should not do any of the sharpness handling if FSR is not enabled.

dlls/winevulkan/vulkan.c
```diff
@@ -3765,6 +4682,47 @@ VkResult WINAPI wine_vkQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR *pP
 
 }
 
+struct wine_vk_sampler_bias
+{
+    float bias;
+    BOOL ansio_only;
```
aeikum: Should this be spelled `aniso_only`?

dlls/winevulkan/vulkan.c
```diff
@@ -1997,84 +2019,871 @@ void main()
...
+#extension GL_GOOGLE_include_directive: require
+
+layout(local_size_x=8, local_size_y=8, local_size_z=1) in;
+
+layout(binding = 0) uniform sampler2D texSampler;
+layout(binding = 1) uniform writeonly image2D outImage;
+
+#define A_GPU 1
+#define A_GLSL 1
+//#include "ffx_a.h"
```
aeikum, line 2085: With the caveat that I am not a graphics dev, it looks weird to have these includes commented out.

aeikum:
Looks cool. I added some nit-picky comments on the review. Some other thoughts:

-Since it touches so much fshack code, which is fragile, this'll need a full QA pass on our end. Just FYI, it may take a little while to get merged. Excited to see those stupid DOOM comments gone, though.

-I'm not a big fan of options. Are all of those environment variables actually useful? There isn't just a good set of hard-coded values we can use?

-If we really need the options, then at the very least these need to be documented. The documentation should include valid values (and accompanying validity checks in code), and what the settings do and why a user would set them. I'm not sure where that documentation would go, maybe a new file in `proton.git/docs/`? But again, I'd prefer if the options just didn't exist.

-How does this interact with the Integer scaling mode, e.g. if both are set? A good solution here may be to provide a new Proton env var which correctly selects between them. E.g. see this (completely untested) diff for a rough concept:

```diff
diff --git a/README.md b/README.md
index 92b008c..32c76da 100644
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ the Wine prefix. Removing the option will revert to the previous behavior.
 | <tt>vkd3dfl12</tt>    |                                | Force the Direct3D 12 feature level to 12, regardless of driver support. |
 | <tt>vkd3dbindlesstb</tt>|                              | Put `force_bindless_texel_buffer` into `VKD3D_CONFIG`. |
 | <tt>hidenvgpu</tt>    | <tt>PROTON_HIDE_NVIDIA_GPU</tt>| Force Nvidia GPUs to always be reported as AMD GPUs. Some games require this if they depend on Windows-only Nvidia driver functionality. See also DXVK's nvapiHack config, which only affects reporting from Direct3D. |
-|                       | <tt>WINE_FULLSCREEN_INTEGER_SCALING</tt> | Enable integer scaling mode, to give sharp pixels when upscaling. |
+|                       | <tt>PROTON_FS_SCALING_MODE</tt>| Use one of the following scaling modes: `integer` scaling to give sharp pixels; AMD's `[fsr](docs/FSR.md)` scaling. |
 | <tt>cmdlineappend:</tt>|                               | Append the string after the colon as an argument to the game command. May be specified more than once. Escape commas and backslashes with a backslash. |
 | <tt>nowritewatch</tt> | <tt>PROTON_NO_WRITE_WATCH</tt> | Disable support for memory write watches in ntdll. This is a very dangerous hack and should only be applied if you have verified that the game can operate without write watches. This improves performance for some very specific games (e.g. CoreRT-based games). |
 | <tt>seccomp</tt>      | <tt>PROTON_USE_SECCOMP</tt>    | **Note: Obsoleted in Proton 5.13.** In older versions, enable seccomp-bpf filter to emulate native syscalls, required for some DRM protections to work. |
diff --git a/proton b/proton
index 82a32e9..f113d7f 100755
--- a/proton
+++ b/proton
@@ -909,6 +909,14 @@ class Session:
         #disable XIM support until libx11 >= 1.7 is widespread
         self.env.setdefault("WINE_ALLOW_XIM", "0")
 
+        if "PROTON_FS_SCALING_MODE" in self.env:
+            if self.env["PROTON_FS_SCALING_MODE"] == "integer":
+                self.env["WINE_FULLSCREEN_INTEGER_SCALING"] = "1"
+                self.env["WINE_FULLSCREEN_FSR"] = "0"
+            elif self.env["PROTON_FS_SCALING_MODE"] == "fsr":
+                self.env["WINE_FULLSCREEN_INTEGER_SCALING"] = "0"
+                self.env["WINE_FULLSCREEN_FSR"] = "1"
+
         if "wined3d11" in self.compat_config:
             self.compat_config.add("wined3d")
 ```

File: dlls/winevulkan/vulkan.c (hunk 7 above)
Comment (DadSchoorse) [reply dlls/winevulkan/vulkan.c:2085]: The build system breaks if I leave them like normal, seems like what ever looks for includes doesn't exclude commented out includes.
```
./../wine/configure
[...]
./../wine/dlls/winevulkan/vulkan.c:2085: error: ffx_a.h: No such file or directory
```

Comment (aeikum) [reply dlls/winevulkan/vulkan.c:2085]: Oh barf. I wonder if we should just move the bytecode stuff to some other file(s)... somehow... anyway whatever, we can just go with what you have if you don't feel like tackling that.

File: dlls/winevulkan/vulkan.c (hunk 1 above)
Comment (DadSchoorse) [reply dlls/winevulkan/vulkan.c:919]: Not really, but it's supported by every driver on all gpus. And we need it to make the compute path conformant. I can check and print a log message I guess.

File: dlls/winevulkan/vulkan.c (hunk 3 above)
Comment (DadSchoorse) [reply dlls/winevulkan/vulkan.c]: Oh I didn't realize I changed that, must have happened when I move some stuff back and forward. Will fix, but as a side note it's really annoying that fshack doesn't follow the typical coding style.

File: dlls/winevulkan/vulkan.c (hunk 1 above)
Comment (aeikum) [reply dlls/winevulkan/vulkan.c:919]: If it's not interesting then I think you can just drop the comment. Up to you, I'm not the right person to make the call.

File: dlls/winevulkan/vulkan.c (hunk 3 above)
Comment (aeikum) [reply dlls/winevulkan/vulkan.c]: Sorry. I'm not a stickler for code style so sometimes I'm bad about following style myself.

File: dlls/winevulkan/vulkan.c (hunk 1 above)
Comment (DadSchoorse) [reply dlls/winevulkan/vulkan.c:919]: I reworded the comment to make clear that this is always supported.

File: dlls/winevulkan/vulkan.c (hunk 2 above)
Comment (DadSchoorse) [reply dlls/winevulkan/vulkan.c:2110]: I replaced all the tabs with the standard 4 spaces

File: dlls/winevulkan/vulkan.c (hunk 6 above)
Comment (DadSchoorse) [reply dlls/winevulkan/vulkan.c]: yes :frog:

File: dlls/winex11.drv/fs.c (hunk 5 above)
Comment (DadSchoorse) [reply dlls/winex11.drv/fs.c]: Reworked it a bit

File: dlls/winevulkan/vulkan.c (hunk 7 above)
Comment (DadSchoorse) [reply dlls/winevulkan/vulkan.c:2085]: I think leaving it like this is the easiest option.

File: dlls/winevulkan/vulkan.c (hunk 3 above)
Comment (DadSchoorse) [reply dlls/winevulkan/vulkan.c]: I think I removed all the accidental changes now.

**Comment by DadSchoorse** [comment]
> -I'm not a big fan of options. Are all of those environment variables actually useful? There isn't just a good set of hard-coded values we can use?
>
>-If we really need the options, then at the very least these need to be documented. The documentation should include valid values (and accompanying validity checks in code), and what the settings do and why a user would set them. I'm not sure where that documentation would go, maybe a new file in proton.git/docs/? But again, I'd prefer if the options just didn't exist.

For sharpness we could leave it at the default, but I would like to have an option since how much sharpening looks good can depend on the blurriness of the input image.

The mip bias stuff has to be entirely optional, it effects everything, even if fsr is not enabled.

> -How does this interact with the Integer scaling mode, e.g. if both are set? A good solution here may be to provide a new Proton env var which correctly selects between them. E.g. see this (completely untested) diff for a rough concept: [...]

Combining integer scaling with fsr would break the effect, yes. I think your idea sounds good.

**Comment by DieFeM** [comment]
I've installed Fedora 34 and luckily I've been able to successfully compile and install proton with your changes (never compiled it before), tested it with Skyrim SE, scaling from 720 to 1080 and it makes it looks like native 1080 (very small differences, almost unnoticeable). This is great, thank you!

PS: I don't understand why this is not yet included as a global filter in the video settings, even if it is not the best implementation, it is way better than most texture filtering options.

**Comment by SkewedZeppelin** [comment]
Anyone who wants an easy way to try this and is already using Steam through Flatpak can:
```
flatpak install --user https://dl.flathub.org/build-repo/52165/com.valvesoftware.Steam.CompatibilityTool.Proton-Exp.flatpakref
```

**Comment by tuxtergames** [comment]
Sorry put a question  here, but my monitor is FHD resolution 1920x1080, to work fsr according amd site I need to left resolution 1477 x  831 to work, is that right?
cause a lot of games I have here don't accept this input resolution, tks all.....

**Comment by DadSchoorse** [comment]
@tuxtergames You would need that in-game resolution to match what AMD calls "Ultra Quality". 1280x720 input would equivalent to "Quality".  And yes, you won't be able to select 1477x831 (at least at the moment) because fshack doesn't expose them as valid modes, see <https://github.com/ValveSoftware/wine/blob/experimental_6.3/dlls/winex11.drv/fs.c#L63> for all currently reported resolutions.

**Comment by tuxtergames** [comment]
> @tuxtergames You would need that in-game resolution to match what AMD calls "Ultra Quality". 1280x720 input would equivalent to "Quality". And yes, you won't be able to select 1477x831 (at least at the moment) because fshack doesn't expose them as valid modes, see https://github.com/ValveSoftware/wine/blob/experimental_6.3/dlls/winex11.drv/fs.c#L63 for all currently reported resolutions.

@DadSchoorse tks for reply xD

**Comment by ren2r** [comment]
The steam overlay stops working when the `WINE_FULLSCREEN_FSR` is enabled. 
When pressing shift-tab on a game running with a 900p resolution on a 1080p screen, the overlay keeps blinking and doesn't shows up.
Removing the FSR flag or reverting the game to the native resolution makes the overlay work again. 
[Edit]
Also, Alt-tab seems to revert to the default scaling method even with the flag enabled.

**Comment by mvrednov** [comment]
Since there is a list of games for which the FSR developers have made support. Will WINE_FULLSCREEN_FSR = 1 and WINE_FULLSCREEN_FSR_STRENGTH keep FSR running for games that have not been supported by the FSR developers ?!

**Comment by Atemu** [comment]
@ren2r I believe I saw something about the ordering of overlays somewhere where the Steam overlay would work if it's at the end or something. Might be worth a Google.

@mvrednov the game would FSR up to its render res (i.e. 720p) and fshack would then FSR that to your native res.

**Comment by rcelyte** [comment]
Does/will this support SteamVR games (i.e. Beat Saber)?

**Comment by Izaic** [comment]
> @tuxtergames You would need that in-game resolution to match what AMD calls "Ultra Quality". 1280x720 input would equivalent to "Quality". And yes, you won't be able to select 1477x831 (at least at the moment) because fshack doesn't expose them as valid modes, see https://github.com/ValveSoftware/wine/blob/experimental_6.3/dlls/winex11.drv/fs.c#L63 for all currently reported resolutions.

Is there something blocking it from going with 1477x831 for Ultra Quality 1080P output FSR?

**Comment by arrowgent** [comment]
ive done some testing with proton-ge-6.14-1 which has amd FSR fullscreen hack
https://github.com/GloriousEggroll/proton-ge-custom/releases/tag/6.14-GE-1
resolutions tested: 1280x720, 1477x831, 1506x847, 1706x960, 1920x1080
monitoring using nvidia-smi and nvtop https://i.imgur.com/r3inDwM.png
comparing steam OSD for fps, and taking a screenshot for comparison between each resolution
i dont use mangohud anymore, it often conflicted with my games, 
libstrangle is not working with proton-ge-6.14, games refuse to launch.  libstrangle works with other proton versions.
vsync is disabled to allow maximum FPS

without question, 1477x831 as reported at https://gpuopen.com/fidelityfx-superresolution/
is about a 98.5:100 quality comparison to 1920x1080
at 1920x1080: gpu usage is 85-95% idling at the same spot (i only have a 1080p monitor, i cant test 1440p or 4k etc)
* 1280x720 does not appear to work in this case, the game is noticeably blurry.  however the fullscreen is upscaled to 1080p.
* with 1477x831: gpu usage drops from 85-90% to 50-59%, FPS increases from 100-107 to 140-147
* similarly 1506x847: i can see the sharpening filter,  gpu usage 60-65%, fps less than above
* finally 1706x960: sharpening filter is noticable over 1920x1080, sometimes it looks much cleaner, less blurry, othertimes appears oversharpened.  gpu usage 73-78% roughly.  fps is less than above.

alternatively in game FPS doesnt go above 77fps in any situation
but the drop in GPU usage makes a difference because at 1080p using 100% gpu i normally get 50-59fps
1477x831 is the biggest difference here, at 100% gpu the FPS doesnt drop below 72fps

intel i7 3940XM & nvidia 1070 mobile 8gb.
game tested: Grim Dawn
the game must report those resolutions in the selections for this to work, if it doesnt report the resolutions it wont work.
visual: https://imgur.com/a/SW8L5hz


other testing: The Long Dark (proton-ge-6.14-1)
this game is not well suited for upscaling techniques, noticable loss in quality in all resolutions: specifically AA
the first person camera also wobbles around constantly, not good for comparison choices.
all game settings are at Ultra

https://imgur.com/a/Sw67f3M

this game is more CPU bound, using 220-250% cpu while running (out of 800% total)
the GPU does benefit by using less SM% at 720p, 831p, 847p, & 960p
no FPS increases occured in any of the above settings, CPU bound with my GPU
going from 960P to 1080p native, a noticable difference in FPS, less CPU% and higher max 100 GPU%
960P = 166-177FPS
1080P = 149-155FPS -- noting the CPU% dropped to average 210-225% while GPU peaked to 95-100%


i will try to think of other games i have for testing... if anyone wants them, im just creating a screenshot/fsr folder and a text file for my observation
**i guess this is just confirmation proof of concept, its working correctly?
also how it works on 1080p monitors, and how the selected resolutions in the game effect performance**

**Comment by mercuriete** [comment]
I did some small testing using proton-ge and I found an issue that I don't know if because of this PR or if because fshack.


I have a 72hz screen (with freesync) and all resolution modes that are native to my screen are shown correctly but some resolutions like 960x540 only show 60hz.

even though freesync should fix the problem, this leads to a issue that when you click on vsync on the game... the game is fixed to 60FPS on a 72hz screen and I would prefer to have the native FPS on the native video mode.


xrandr output on my main monitor
```
Screen 0: minimum 320 x 200, current 3840 x 1080, maximum 16384 x 16384
DisplayPort-0 connected primary 1920x1080+0+0 (normal left inverted right x axis y axis) 598mm x 336mm
   1920x1080     71.91*+  60.00    59.94  
   1680x1050     59.95  
   1600x900      60.00  
   1280x1024     60.02  
   1440x900      59.89  
   1280x800      59.81  
   1280x720      60.00    59.94  
   1024x768      70.07    60.00  
   800x600       72.19    60.32    56.25  
   640x480       72.81    66.67    60.00    59.94  
   720x400       70.08  
```

PS: In my case the games are properly rendered and it looks great this issue is only about refresh rate.

**Comment by arrowgent** [comment]
indeed that does seem to be partly hardcoded into proton-ge's fshack, _currently_.

another issue is dual monitor setups,  i can push a screen left or right, but FShack only works on the "main" "left" "display:0" monitor starting at 0x0, if i push it to the right @ 1921x0 (which is my 120/144hz monitor) _the game cannot be interacted with correctly._

**Comment by arrowgent** [comment]
other tests:

**Road Redemption** no changes in FPS, resolution selections are present 
960 x  540, 1129 x  635, 1280 x  720, 1477 x  831, 1506 x  847, 1706 x  960, & 1920 x 1080
obvious loss in visual quality
minor improvements in GPU usage SM%

**warfork** https://imgur.com/a/N2uAQlP

**rust** https://imgur.com/a/0YBk8P9
**rust** https://imgur.com/a/upQY3yR
960 x  540, 1129 x  635, 1280 x  720, 1477 x  831, 1506 x  847, 1706 x  960, & 1920 x 1080

i was also able to fshack 640x480 & 720x400 aspect correct with a fullscreen window of 1920x1080, which is cool.
https://i.imgur.com/hcCoGhj.png


either the FSR isnt working correctly, or the game isnt allowing scaling?
i can see the GPU SM% reduction, but my CPU usage% stays the same
no FPS increases or decreases noticable on these.
again suggesting possibly a CPU bottleneck

while the _fullscreen upscale hack works and creates a 1080p window_ for all these resolutions (not always) it doesnt seem to be doing anything else

**Comment by arrowgent** [comment]
in other testing,

im thinking the **Unity engine** has no performance benefit from FSR unless your GPU usage% is already at 100%
possibly how Unity creates a _Fullscreen_ window
and the WINE_FULLSCREEN_FSR_STRENGTH does not actually sharpen the rendering image

i was only able to test one **Unreal Engine** game and saw no performance benefit or visual difference on screen. 
the fshack was not able to correct the window size between Fullscreen & borderless fullscreen,  assuming Fullscreen was not correctly working with fshack and borderless just created a window within the native resolution (smaller window in window)
i'll need to look at my demos to see if i have other UE4 games

by comparison, **Grim Dawn** does seem to have noticable performance gain with this and FSR sharpening & scaling seem to work great with fshack
noted Grim Dawn does not use Unity or Unreal engines.

i should test out id-tech engine,  possibly **Doom3** or similar.   i'll need to create a backup of my save data before using proton-ge-6.14-1


`WINE_FULLSCREEN_FSR_STRENGTH` i now tested with =0 =1 =2 and =5
its difficult to see if anything significant is different here, i dont take enough screenshots!
since no visible impact on cpu/gpu im leaving it at =0 for maximum sharpening effect

**Comment by mercuriete** [comment]
@arrowgent if you have an amdgpu you should see power consumption in sensors.
I recommend you to put ksysguard on the secondary screen monitoring power consumption and do the test using vsync enabled.

For example, on red read redemption 2 my GPU goes from 100W to 75W using this PR.

Percentage of GPU usage is not reliable when your GPU clockdown to minimum MHz. It will show 100% when in reality is doing nothing.

**Comment by arrowgent** [comment]
nvidia, i was using nvtop & nvidia-smi
https://i.imgur.com/r3inDwM.png
so i can see SM% used, mhz, & power wattage, etc...

i should use DXVK HUD more for testing,  it kinda gets in the way but i did some minor tests with it.  not much difference compared to using steam's overlay for FPS

i dont really see too many other people doing testing reports for wine/proton FSR, which is why i gave some results...

**Comment by Zeioth** [comment]
May this work in borderless mode in the future? Or is Fullscreen a limitation of FSR?

**Comment by kode54** [comment]
> May this work in borderless mode in the future? Or is Fullscreen a limitation of FSR?

It needs some way for the game resolution to be different from the display resolution, so it only works with the full screen hack, which replaces actual mode switching with upscaling to the desktop resolution.

It stands to reason it could also be used for upscaling with Gamescope, if implemented into that compositor. Doing it for arbitrary windows would require hooking the exact display window and reporting the client visible area as a fraction of the actual size.

**Comment by pchome** [comment]
Well, I patched FS-hack code to add "cinematic" ultrawide (1080p) resolutions, but I can see how quickly `fs_monitor_sizes[]` can go brr. So ...
> * An environment variable to add custom resolution modes would be nice to more directly control the quality. e.g. 2048x1152

maybe `WINE_FULLSCREEN_FSR` environment variable should accept scale factor? According to FSR info page it should work with scale factors from 1 to 4 (presets are 1.3x/1.5x/1.7x/2x from Ultra down to Performance).
> In addition to fixed scaling, FSR may be used in “arbitrary scaling” mode, whereby any area scale factor between 1x and 4x is supported. This mode is typically used for Dynamic Resolution Scaling, whereby source resolution is determined by a fixed performance budget to achieve a minimum frame rate.

POC patch, but w/ additional env var:
```patch
--- a/dlls/winex11.drv/fs.c
+++ b/dlls/winex11.drv/fs.c
@@ -177,6 +177,8 @@ static void add_fs_mode(struct fs_monito
 static BOOL fs_monitor_add_modes(struct fs_monitor *fs_monitor)
 {
     DEVMODEW *real_modes, *real_mode, current_mode;
+    const char *fsr_scale_env;
+    double fsr_scale_factor;
     UINT real_mode_count;
     DWORD width, height;
     ULONG_PTR real_id;
@@ -203,6 +205,22 @@ static BOOL fs_monitor_add_modes(struct
         return FALSE;
     }
 
+    /* For FSR add one mode in range 1x up to 4x */
+    if ((fsr_scale_env = getenv("WINE_FULLSCREEN_FSR_SCALE")))
+    {
+        fsr_scale_factor = atof(fsr_scale_env);
+        if (fsr_scale_factor > 1 && fsr_scale_factor <= 4)
+        {
+            add_fs_mode(fs_monitor, current_mode.dmBitsPerPel,
+                (UINT)ceil(current_mode.dmPelsWidth / fsr_scale_factor),
+                (UINT)ceil(current_mode.dmPelsHeight / fsr_scale_factor),
+                current_mode.dmDisplayFrequency,
+                current_mode.u1.s2.dmDisplayOrientation);
+
+            real_settings_handler.free_modes(real_modes);
+            return TRUE;
+        }
+    }
     /* Add the current mode early, in case we have to limit */
     add_fs_mode(fs_monitor, current_mode.dmBitsPerPel, current_mode.dmPelsWidth,
                 current_mode.dmPelsHeight, current_mode.dmDisplayFrequency,
```
#### NOTE: for users who will try this patch, this will add only one supported resolution and will likely fail for games which expect some standard modes to be present.

EDIT: My original goal was to limit available modes when FSR enabled to only "presets" and native/current resolution. But I tested a few random games w/ this patch and didn't even touched in-game configuration. They started in desired mode and all I had to do is start new game. I sometimes see some Wine errors in terminal output, but that was expected.

EDIT2: I removed the screenshot, appeared not only shadows were broken but also FSR. Anyway it was for demonstration that WINE_FULLSCREEN_FSR_SCALE=4 work, image resolution was 2560x1080 while 640x270 was used by the benchmark.

DadSchoorse:
Closing this since I don't think it will get merged given the limitations. If you want FSR I think https://github.com/Plagman/gamescope/pull/360 is the better option since it will work with every game.

gabriele2000:
> Closing this since I don't think it will get merged given the limitations. If you want FSR I think [Plagman/gamescope#360](https://github.com/Plagman/gamescope/pull/360) is the better option since it will work with every game.

What about RSR? Same tech, driver-integrated, perhaps it could be integrated in mesa, to make it global even for native linux games?

DadSchoorse:
Well afaiu RSR is exactly that, FSR without per game integration. No need to put it in the driver, the correct place to do upscaling is the compositor. And you should be able to use gamescope's FSR with all linux games, even native.

Atemu:
I agree that the compositor is the best place for this but having it in fshack is still useful to those who cannot use gamescope for one reason or another.

DadSchoorse:
@Atemu feel free to keep using this code then, I think TKG/GE still maintain a rebased version of it.

Atemu:
That's pretty much the status quo anyways. The point of all of this is upstream Proton inclusion of the feature though which can only happen here.

DadSchoorse:
At this point it's pretty clear that it won't be accepted into upstream Proton.

mercuriete:
I think a problem here (please correct me if I am wrong) is that gamescope is Wayland only.
There are some users that just because they wanted to use X11 or any non-justifiable reason they don't want to migrate to Wayland.
Given said that... if gamescope becomes popular I will move to Wayland in the long run.

DadSchoorse:
Gamescope works just fine in nested mode on X11. The real (hopefully temporary) problem is that it does not work on nvidia because they don't support some required system integration extensions.

aeikum:
> At this point it's pretty clear that it won't be accepted into upstream Proton.

FWIW the only outstanding issues from my review above were how best to handle and document the environment variable options. But, it sounds like it's more appropriate if it lives at a lower level than Proton itself anyhow.

pchome:
I wonder if its possible to move Vulkan part (so all tuning) to vulkan layer. Then Wine can do it's part if such layer was loaded. All documentation will be on layer's side. Maybe even implementation (FSR, NVIDIA's thing, whatever else will be invented) could be easily switched by loading different vulkan layer.

For example if an implicit layer launching by BLAHBLAH_FSR=1 environment variable then it can be added to Steam command options like `BLAHBLAH_FSR=1 %command%` (iirc). Wine will check that env var or loaded layer name or else detection mechanics, maybe will read some additionally required data from environment or from layer (iirc layer can provide such data).
Also such layers could be used by other applications or compositors, they will need to do their  part with window (if I understand this correctly).

kakra:
> Gamescope works just fine in nested mode on X11. The real (hopefully temporary) problem is that it does not work on nvidia because they don't support some required system integration extensions.

Yeah, I tried and it trips over unimplemented buffer modifiers with NVIDIA:
> vulkan: selecting physical device 'NVIDIA GeForce GTX 1660 Ti'
> vulkan: physical device does not support DRM format modifiers
> vulkan: vkCreateDevice failed (VkResult: -7)

But what does it mean for wayland? I think this error won't go away when we finally could use wayland on NVIDIA just fine? (given that some outstanding problems in desktop environment integration became fixed)

DadSchoorse:
Nvidia needs to implement `VK_EXT_external_memory_dma_buf` and `VK_EXT_image_drm_format_modifier`. Don't ask me why they haven't done that yet, given that they support similar functionality with gbm+OpenGL, which most other wayland compositors use.

arrowgent:
for anyone that continues to monitor the progress of wine/proton AMD FSR

please see community patches here:

https://github.com/Frogging-Family/community-patches/blob/master/wine-tkg-git/amd_fsr_fshack.mypatch

if you know of other community maintained patches, post em up!


thanks a ton for dadschoorse for his contribution

PR #118, opened 2021-07-26
Add disable fs hack option

yuiiio:
add WINE_DISABLE_FULLSCREEN_HACK env
It is useful for testing.

### DIFF
--- a/dlls/winex11.drv/vulkan.c
+++ b/dlls/winex11.drv/vulkan.c
@@ -825,6 +825,18 @@ static VkSurfaceKHR X11DRV_wine_get_native_surface(VkSurfaceKHR surface)
     return x11_surface->surface;
 }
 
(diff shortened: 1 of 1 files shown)


aeikum:
How have you found this useful?

yuiiio:
I thought it would be useful to be able to explicitly set the other side when the game or compositor is doing its own scaling.

If it matches the resolution, it's disabled, so it may not be necessary.
If it doesn't match the resolution, the input will be weird, so it's probably best not to add it.

When I think about it, I don't really need it, so I'll close it.

kattjevfel:
Wait what, this is a possibility? I'd love this feature to fix using Wine virtual desktop in Proton! Currently it's all kinds of broken with fshack and no effort is put into that, while this would provide an instant fix. see https://github.com/ValveSoftware/Proton/issues/4673

PR #119, opened 2021-08-18
Add nvcuda thunk implementation for NVIDIA DLSS

liam-middlebrook:
This is needed for DLSS versions 2.1.38 through 2.1.40 to work within
Proton, as they contain a link-time dependency on nvcuda.dll. In the
case of what is needed by DLSS, no ordinals need to be exported for
nvcuda.dll to successfully load.


ivyl:
this is included in experimental https://github.com/ValveSoftware/wine/commits/experimental-wine-6.3-20210916 and is enabled with nvapi https://github.com/ValveSoftware/Proton/commit/708cd8d22f7343526bcf272338dd4aad1a2d4aab

thanks!

PR #123, opened 2021-09-05
wine.inf: Use native xinput1_3 for WRC10.

whizse:
Hi,

Here's a trivial PR to set xinput1_3 to native for the game WRC10 too. 

It has the same issues as WRC8 discussed in #121, with the added problem that it defaults to, and crashes with, DX12. See issue [#5143.](https://github.com/ValveSoftware/Proton/issues/5143)


ivyl:
Thanks! This should be in the next release of experimental. It looks like the DX12 issues were recently fixed, so they should be included with the next vkd3d-proton bump.

I'll squash all the overrides (including one for the upcoming WRC 9) into a single commit when it comes to the non-experimental release.

whizse:
Awesome! Thanks!

ivyl:
This is now live in experimental. VKD3D-Proton also has been updated to the newest master so the game should be playable.

I think we can close this pull request.

kisak-valve:
Commit found at https://github.com/ValveSoftware/wine/commit/7e69a5b46325e5ac12e19a91f8eba7375e6011f3.

PR #125, opened 2021-09-26
kernelbase: Don't pass StdHandles with CREATE_NEW_CONSOLE.

redmcg:
Wine-Bug: https://bugs.winehq.org/show_bug.cgi?id=51264

This patch (accepted to Wine 6.12) fixes an annoying (but harmless) crash in the Elite Dangerous launcher.

I'm not sure if experimental is the right branch to raise a PR against; so please let me know if I should be raising against another.

Signed-off-by: Brendan McGrath <brendan@redmandi.com>
Signed-off-by: Jacek Caban <jacek@codeweavers.com>
Signed-off-by: Alexandre Julliard <julliard@winehq.org>


ivyl:
I've merged this commit and it will be a part of the next experimental release. Thanks!

redmcg:
Awesome. Thank-you!

PR #128, opened 2021-10-19
ntdll/fsync: Support only futex_waitv() API

andrealmeid:
Since an interface for wait on multiple futexes was merged, we no longer need to support different interfaces. Drop out FUTEX_WAIT_MULTIPLE (opcode 31) and old futex2 interface (with futex2_wake/wait) in favor of the interface accepted by upstream Linux.

If an user uses Proton with a kernel with a different futex interface, it's unsupported and the Wine will fallback to use esync. The futex_waitv() interface will be released at Linux 5.16, but users can apply backports to custom kernels with patches available here: https://github.com/andrealmeid/futex_waitv_patches

### DIFF
--- a/dlls/ntdll/unix/fsync.c
+++ b/dlls/ntdll/unix/fsync.c
@@ -76,142 +76,23 @@ static inline void small_pause(void)
 #endif
 }
 
-static LONGLONG update_timeout( ULONGLONG end )
+static inline int futex_wait_multiple( const struct futex_wait_block *futexes,
+        int count, const struct timespec *timeout )
 {
-    LARGE_INTEGER now;
-    LONGLONG timeleft;
-
-    NtQuerySystemTime( &now );
-    timeleft = end - now.QuadPart;
-    if (timeleft < 0) timeleft = 0;
-    return timeleft;
+    return syscall( __NR_futex, futexes, 31, count, timeout, 0, 0 );
 }
 
-/* futex2 experimental interface */
-
-static long nr_futex2_wait, nr_futex2_waitv, nr_futex2_wake;
-
-#define FUTEX_32 2
-#define FUTEX_SHARED_FLAG 8
-#define FUTEX_CLOCK_REALTIME 256
-
-struct futex_waitv
+static inline int futex_wake( int *addr, int val )
 {
-    void *uaddr;
-    unsigned int val;
-    unsigned int flags;
-};
+    return syscall( __NR_futex, addr, 1, val, NULL, 0, 0 );
+}
+
+static inline int futex_wait( int *addr, int val, struct timespec *timeout )
+{
+    return syscall( __NR_futex, addr, 0, val, timeout, 0, 0 );
+}
 
 static unsigned int spincount = 100;
-
-static inline int futex_wake( int *addr, int count )
-{
-    if (nr_futex2_wake)
-        return syscall( nr_futex2_wake, addr, count, FUTEX_32 | FUTEX_SHARED_FLAG );
-    return syscall( __NR_futex, addr, 1, count, NULL, 0, 0 );
-}
-
-struct timespec64
-{
-    long long tv_sec;
-    long long tv_nsec;
-};
-
-static inline int futex_wait( int *addr, int val, const ULONGLONG *end )
-{
-    if (nr_futex2_wait)
-    {
-        if (end)
-        {
-            struct timespec64 timespec;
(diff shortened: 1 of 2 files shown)


kisak-valve:
Hello @andrealmeid, we'll see if a Proton dev is happy with this as-is, but it might make sense to first revert commits like https://github.com/ValveSoftware/wine/commit/9fa46d16a398832a37ee6c2b932dbf9d3e3cee09 which didn't land in Proton 6.3 and go from there, or maybe it makes sense to cut more out. The idea being to reduce the total amount of commits that need to be evaluated in major rebases going forward. (This is a non-authoritative suggestion.)

andrealmeid:
Hi @kisak-valve, indeed that makes more sense, I'll redo this to have a commit that can be applied in both experimental and 6.3, thanks for pointing that out.

zfigura:
The patch looks correct, but I'm going to add a few nitpicks for maintainability.

```
+/* futex_waitv interface */
+
+#define FUTEX_32 2
+#ifndef __NR_futex_waitv
+#define __NR_futex_waitv 449
+#endif
+
+#define u64_to_ptr(x) (void *)(uintptr_t)(x)
+
+struct futex_waitv {
+    uint64_t   val;
+    uint64_t   uaddr;
+    uint32_t   flags;
+    uint32_t __reserved;
+};
```

Shouldn't FUTEX_32 and futex_waitv also be guarded by an #ifndef? I guess you could use the same __NR_futex_waitv definition.

```
+static unsigned int spincount = 100;
```

This seems to have been moved unnecessarily.

```
+static inline void futex_vector_set( struct futex_waitv *waitv, unsigned int index, int *addr, int val )
+{
+    waitv[index].uaddr = (uintptr_t) addr;
+    waitv[index].val = val;
+    waitv[index].flags = FUTEX_32;
+    waitv[index].__reserved = 0;
+}
```

I would prefer to pass a single pointer to this function, i.e. "futex_vector_set( &futexes[i], addr, val )".

```
-                ret = futex_wait_multiple( futexes, waitcount, &tmo_p );
+                if (waitcount == 1)
+                    ret = futex_wait( u64_to_ptr(futexes[0].uaddr), futexes[0].val, &end );
+                else
+                    ret = futex_wait_multiple( futexes, waitcount, &end );
```

Can this be handled in futex_wait_multiple() instead?

For that matter, do we really want to do this, especially if there's a timeout? Note that futex_wait() will end up doing an extra (v)syscall. Have you measured a performance benefit?

```
-static inline int futex_wake( int *addr, int val )
+static inline int futex_wake( int *addr, int count )
 {
-    return syscall( __NR_futex, addr, 1, val, NULL, 0, 0 );
+    return syscall( __NR_futex, addr, 1, count, NULL, 0, 0 );
 }
```

This change seems unnecessary. Yes, it's more correct, but I'd like to avoid touching more code than we need to; rebases are painful enough already.

(EDIT: code formatting)

andrealmeid:
Hi @zfigura, thanks for the feedback. I pushed a work in progress branch, sorry for the small mistakes. I now pushed a complete work incorporating your feedback, please have a look.

> Can this be handled in futex_wait_multiple() instead?
> For that matter, do we really want to do this, especially if there's a timeout? Note that futex_wait() will end up doing an extra (v)syscall. Have you measured a performance benefit?

Yes, `futex_wait_multiple()` can do it as well. I benchmarked for the case with no timeout and I could measure that `futex_wait()` is noticeable faster than `futex_wait_multiple()` given that it needs an extra user copy in the kernel side.

About the timeout path, I changed the `futex_wait()` call to use absolute timeouts as well, so it doesn't need to use the additional syscall anymore.

zfigura:
Okay, I think that looks correct, at least from visual inspection. Thanks, looks good.

aeikum:
@zfigura merged this into 6.3 and Experimental. Should be released in Experimental soon and in the next 6.3 release.

ivyl:
Hi. Those changes cause 100% CPU utilization when running on a kernel with old version of the futex2 patches. There's quite a few users of downstream kernels that carry those patches. It would be nice if this was considered and tested for.

We need something like: https://gist.github.com/ivyl/93107ffef32b2e8816d1176286afdd12

ivyl:
Those patches + the change that detects the old futex2 patches and disables fsync have been included in the latest experimental.

andrealmeid:
Hi @ivyl, thanks for reporting and patching this. I had tested it, but with an old futex2 patchset versions that hadn't 449 as a futex syscall, so that's why I didn't see the error then, my bad.

PR #139, opened 2022-02-23
winebus.sys: Fix container detection

redmcg:
msvcrt 'access' will return 0 if the file is present and readable

### DIFF
--- a/dlls/winebus.sys/bus_udev.c
+++ b/dlls/winebus.sys/bus_udev.c
@@ -2098,7 +2098,7 @@ NTSTATUS udev_bus_init(void *args)
         goto error;
     }
 
-    if (access("/run/pressure-vessel", R_OK) || access("/.flatpak-info", R_OK))
+    if (access("/run/pressure-vessel", R_OK) == 0 || access("/.flatpak-info", R_OK) == 0)
     {
         TRACE("Container detected, bypassing udevd by default\n");
         options.disable_udevd = TRUE;


ivyl:
Nice catch.

FWIW this is not msvcrt's `access` but glibc's one. The code lives on the unix side.

redmcg:
> FWIW this is not msvcrt's access but glibc's one. The code lives on the unix side.

Oops, of course; not sure what I was thinking.

A question unrelated to this PR: is there a way to disable the container when launching a game in Proton? I was just searching the code to see if there was an env variable or something that could be set - but you might know off the top of your head.

ivyl:
On Arch Linux there's a package named `steam-native` that install all the dependencies for Steam itself and runs things without the runtime:

```
$ cat $(which steam-native)
#!/bin/bash
export STEAM_RUNTIME=0
export STEAM_RUNTIME_HEAVY=0
# Workaround for dbus fatal termination related coredumps (SIGABRT)
# https://github.com/ValveSoftware/steam-for-linux/issues/4464
export DBUS_FATAL_WARNINGS=0
# Override some libraries as these are what games linked against.
export LD_LIBRARY_PATH="/usr/lib/steam:/usr/lib32/steam${LD_LIBRARY_PATH:+:}$LD_LIBRARY_PATH"
exec /usr/lib/steam/steam "$@"
```

It may do what you want. Just keep in mind that running Proton outside of the provided containers is unsupported.

redmcg:
Awesome - thanks for that - I'll take a look.

PR #140, opened 2022-02-23
kernelbase: HACK: Force swiftshader + in-process-gpu for Rockstar Soc…

misyltoad:
…ial Club

'Fixes' it being one frame behind in all of its rendering.

Using WineD3D etc, doesn't help here either, but this works for now.


ivyl:
this is live in bleeding edge and should be included in the next experimental release

PR #141, opened 2022-03-31
winevulkan: Support for NVIDIA Image Scaling

esullivan-nvidia:
This pull request adds an option to fshack to upscale applications to the current display resolution with NVIDIA Image Scaling. It can be enabled by setting the WINE_FULLSCREEN_UPSCALER environment variable to "NIS". Sharpness can be adjusted with WINE_NIS_UPSCALER_SHARPNESS which can be set between [0, 100] with 100 being the maximum sharpness value. The NIS upscaler also supports using FP16 storage by default but it will fallback to using FP32 if FP16 is not supported or WINE_NIS_UPSCALER_FORCE_FP32 is set to 1.


Calinou:
Is there a sharpness adjustment available somewhere (like in the Windows driver)? FSR has a similar adjustment available with an environment variable: https://linux-gaming.kwindu.eu/index.php?title=FSR_-_FidelityFX_Super_Resolution

Also, it might be worth using some sharpening by default (33% is recommended in the Windows driver).

esullivan-nvidia:
> Is there a sharpness adjustment available somewhere (like in the Windows driver)? FSR has a similar adjustment available with an environment variable.
> 
> Also, it might be worth using some sharpening by default (33% is recommended in the Windows driver).

I have updated the description to specify the environment variable that sets the sharpness.

dlls/winevulkan/vulkan.c
```diff
@@ -2473,11 +2211,20 @@ NTSTATUS wine_vkCreateSwapchainKHR(void *args)
...
+        }
+
         if(native_info.imageFormat != VK_FORMAT_B8G8R8A8_UNORM &&
                 native_info.imageFormat != VK_FORMAT_B8G8R8A8_SRGB){
             FIXME("swapchain image format is not BGRA8 UNORM/SRGB. Things may go badly. %d\n", native_info.imageFormat);
         }
 
+        device->phys_dev->instance->funcs.p_vkGetPhysicalDeviceFormatProperties(device->phys_dev->phys_dev, native_info.imageFormat, &formatProperties);
+        if(!(formatProperties.optimalTilingFeatures & VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT)){
+            native_info.imageFormat = VK_FORMAT_B8G8R8A8_UNORM;
```
DadSchoorse, line 2225: This breaks SRGB swapchains if the application uses it in vkCmdBlitImage. It will also cause validation errors for image views that the application creates, but that fine I guess because no driver actually cares afaik.

libs/nis/NIS/NIS_Main.glsl
```diff
@@ -0,0 +1,93 @@
...
+    uint kOutputViewportWidth;
+    uint kOutputViewportHeight;
+
+    float reserved0;
+    float reserved1;
+};
+
+layout(set=0,binding=1) uniform sampler samplerLinearClamp;
+layout(set=0,binding=2) uniform texture2D in_texture;
+layout(set=0,binding=3,rgba8_snorm) uniform image2D out_texture;
```
DadSchoorse: You want `writeonly` here to not cause problems on intel. Also this is not an rgba8_snorm image.

dlls/winevulkan/nis.c
```diff
@@ -0,0 +1,1365 @@
...
+    nis_data->use_fp16 = fp16 && strcmp(fp16, "0");
+
+    device->phys_dev->instance->funcs.p_vkGetPhysicalDeviceFormatProperties(device->phys_dev->phys_dev, VK_FORMAT_R16G16B16A16_SFLOAT, &formatProperties);
+    if (!(formatProperties.optimalTilingFeatures & VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT)){
+        WARN("FP16 coefficients are enabled but the device does not support using them as a sampled image, falling back to FP32\n");
+        nis_data->use_fp16 = FALSE;
+    }
+
+    sharpness = getenv("WINE_NIS_UPSCALER_SHARPNESS");
+    nis_data->sharpness = sharpness != NULL ? atof(sharpness) : 0.5f;
```
aeikum: Should have sanity checking here, both for valid range and string-to-float conversion failure.

dlls/winevulkan/nis.c
```diff
@@ -0,0 +1,1365 @@
...
+
+    swapchain->upscaler_data = nis_data;
+
+    /* Get the NIS upscaler options */
+    fp16 = getenv("WINE_NIS_UPSCALER_USE_FP16");
+    nis_data->use_fp16 = fp16 && strcmp(fp16, "0");
+
+    device->phys_dev->instance->funcs.p_vkGetPhysicalDeviceFormatProperties(device->phys_dev->phys_dev, VK_FORMAT_R16G16B16A16_SFLOAT, &formatProperties);
+    if (!(formatProperties.optimalTilingFeatures & VK_FORMAT_FEATURE_SAMPLED_IMAGE_BIT)){
+        WARN("FP16 coefficients are enabled but the device does not support using them as a sampled image, falling back to FP32\n");
```
aeikum: This warning seems to disagree with the logic. Does the condition need `if (use_fp16)`, too?

aeikum:
What do `WINE_NIS_UPSCALER_SHARPNESS` and `WINE_NIS_UPSCALER_USE_FP16` do, and when would a user want to set them? Is it possible to do the "right thing" for a given situation without requiring user intervention? If we really need those variables, then they should be documented someplace. That could live in the Proton repo's `docs` directory.

We'll also need to update docs for the new `WINE_FULLSCREEN_UPSCALER` variable.

What happens if this is enabled on non-nvidia hardware? If it will crash, we should detect that and refuse to use it, or at least print a very visible error message. If it will just not work, then we should at least print a warning.

dlls/winex11.drv/fs.c
```diff
@@ -535,6 +535,26 @@ BOOL fs_hack_is_integer(void)
...
+    static enum fs_hack_upscaler upscaler = X11_DRV_UPSCALER_NONE;
+
+    if (upscaler == X11_DRV_UPSCALER_NONE)
+    {
+        const char *e = getenv("WINE_FULLSCREEN_UPSCALER");
+        if (e && !strcmp(e, "NIS"))
+        {
+            upscaler = X11_DRV_UPSCALER_NIS;
+        }
+        else
```
aeikum, line 553: Maybe add a warning here for unrecognized values?

dlls/winevulkan/nis.c
esullivan-nvidia: Yes, that should be part of the condition here as well. Good catch.

aeikum:
There's a wider conversation of if Proton is the correct place for this scaling to live. The FSR code ended up being integrated into gamescope, instead of Proton. Should NIS maybe live there, as well? I'm not the right person to answer this, but wanted to start the conversation before you go too much further in Proton.

Calinou:
> What do `WINE_NIS_UPSCALER_SHARPNESS` and `WINE_NIS_UPSCALER_USE_FP16` do, and when would a user want to set them? Is it possible to do the "right thing" for a given situation without requiring user intervention? If we really need those variables, then they should be documented someplace. That could live in the Proton repo's docs directory.

Sharpness adjustment is often subjective – some people will prefer more aggressive sharpening, while others dislike any form of sharpening. Sharpening also looks better on some games than it does on others.

On the other hand, I think FP16 usage should be automatic whenever it's supported (and disabled when it's not). The quality difference is usually not noticeable during actual gameplay, but the performance boost definitely is.

pchome:
Also external sharpeners exist which gamers already familiar with. Like CAS (via vkbasalt) or `__GL_SHARPEN_ENABLE` in NVIDIA drivers. So I guess there should be "tips and tricks" documentation section which will tell user how to improve image quality when upscaler is enabled. In particular about anti-aliasing, which is "required" for FSR to do it's best (I guess this rule is common for upscalers).

P.S. My personal preferred sharpness settings: casSharpness=0 or __GL_SHARPEN_VALUE=20, which give almost similar minimal sharpening. Enabled by default for almost all games (some games have built-in sharpening).

arrowgent:
> There's a wider conversation of if Proton is the correct place for this scaling to live. The FSR code ended up being integrated into gamescope, instead of Proton. Should NIS maybe live there, as well? I'm not the right person to answer this, but wanted to start the conversation before you go too much further in Proton.

yeah im curious why the jump to valve proton instead of upstream mainline wine/wineHQ

the AMD FSR now lives as a community wine patch for builders as an optional compilation
https://github.com/Frogging-Family/community-patches/tree/master/wine-tkg-git

aeikum:
The question isn't really "upstream vs proton," it's more "wine vs something further down the stack."  For example if it lives in gamescope, then native games can benefit, too.

libs/nis/NIS/NIS_Main.glsl
esullivan-nvidia: Fixed in this PR and the NIS repository.

dlls/winevulkan/nis.c
esullivan-nvidia: I updated WINE_NIS_UPSCALER_SHARPNESS to accept an integer value from 0 to 100 rather than a float. Also added a check to confirm that the input is an integer.

dlls/winevulkan/nis.c
esullivan-nvidia: I have made it so FP16 is the default mode if the device supports it and the user has not enabled FP32 usage.

dlls/winex11.drv/fs.c
esullivan-nvidia: Done.

ivyl:
Hi. Thanks for the PR! IMO adding all this complexity here and diverging even further from upstream Wine is not a good direction.

Gamescope is the best place for features like this - it allows running on bare metal but also in a nested session in X11 / other Wayland compositors. Additionally it exposes runtime options to control the upscaler. Also all the APIs, not only things built atop Vulkan, will benefit from it - OpenGL, older DirectXes, etc.

`gamescope $OPTIONS -- %command%` is about as complex as `WINE_FULLSCREEN_UPSCALER=... WINE_UPSCALER_OPTIONS=... %command%`. 

My understanding is that there's an ongoing effort to implement the missing Vulkan extensions so gamescope can also run on Nvidia's HW. It also already has FSR support. If NIS provides any benefits over it I would try to contribute there instead.

esullivan-nvidia:
Thanks for the feedback everyone, I really appreciate it! I have opened a pull request that adds NIS into Gamescope and am going to go ahead and close this PR.

https://github.com/Plagman/gamescope/pull/488

PR #148, opened 2022-05-13
winevulkan: Fix unwrapping VK_NULL_HANDLE.

DadSchoorse:
Fixes unbinding sparse memory.
Not needed in upstream/stable Proton because those don't wrap VkDeviceMemory


ivyl:
Would you mind if this got squashed into 325559b5a255666e41493bd9e307fe4a9d79e801 which introduces the wrapping?

DadSchoorse:
Not at all.

DadSchoorse:
Looks like the fix is now in squashed into experimental, so I'll close this.

PR #155, opened 2022-07-22
qedit: Do not force the sample to be freed in SampleGrabber_callback().

redmcg:
This patch removes a work around that causes a crash in Unravel Two. [#3962](https://github.com/ValveSoftware/Proton/issues/3962)

There is a callback in Unravel Two that appears to add a reference to a
IMediaSample, which this workaround treats as a leak and releases. However, the
application also later releases the reference itself, causing a use-after-free.

Note that this patch was recently accepted upstream to Wine.

### DIFF
--- a/dlls/qedit/samplegrabber.c
+++ b/dlls/qedit/samplegrabber.c
@@ -164,9 +164,6 @@ static void SampleGrabber_callback(struct sample_grabber *This, IMediaSample *sa
 		if (ref)
 		{
 		    ERR("(%p) Callback referenced sample %p by %lu\n", This, sample, ref);
-		    /* ugly as hell but some apps are sooo buggy */
-		    while (ref--)
-			IMediaSample_Release(sample);
 		}
 	    }
             break;


kisak-valve:
Upstream commit for reference: https://gitlab.winehq.org/wine/wine/-/commit/1113f42aa748d9e0a27c4663b9d8de1237f95e73

ivyl:
Hi, thanks for the fix! This is now merged into our staging experimental branch and is already in bleeding-edge.

https://github.com/ValveSoftware/wine/commits/experimental-wine-bleeding-edge-7.0-20567-20220722-p23b574-wc88dca-d5aa943-vbe2aaf

When you are cherry-picking commits please use `-x` switch so it records the upstream commit in the message. This helps a lot when doing a rebase onto a newer Wine. You can add a note on any modifications that had to be done, like not including the test changes. See the modified commit under the bleeding-edge tag for reference :-)

redmcg:
> This is now merged into our staging experimental branch

Awesome, thank-you. Should I close this PR?

> When you are cherry-picking commits please use `-x` switch so it records the upstream commit in the message.

Got it. I'll keep that in mind for next time. Thanks again.

PR #158, opened 2022-08-26
winevulkan: fshack: Fixes and improvements.

DadSchoorse:
Old fshack causes radv assertions because of image view format mismatches.

Changes:
- drop the vkCmdBlitImage path, it's impossible to keep without
  VK_KHR_swapchain_mutable_format.
- drop the compute+copy path, all drivers support swapchains with storage usage,
  no reason to keep untested code around.
- interpolate in linear color space for slightly better results
- improve compute path by interpolating at the pixel center
- fix all validation errors

### DIFF
--- a/dlls/winevulkan/vulkan.c
+++ b/dlls/winevulkan/vulkan.c
@@ -803,6 +803,8 @@ VkResult WINAPI __wine_create_vk_device_with_callback(VkPhysicalDevice phys_dev,
         VkResult (WINAPI *native_vkCreateDevice)(VkPhysicalDevice, const VkDeviceCreateInfo *, const VkAllocationCallbacks *,
         VkDevice *, void * (*)(VkInstance, const char *), void *), void *native_vkCreateDevice_context)
 {
+    VkPhysicalDeviceFeatures features = {0};
+    VkPhysicalDeviceFeatures2 *features2;
     VkDeviceCreateInfo create_info_host;
     struct VkQueue_T *next_queue;
     struct VkDevice_T *object;
@@ -835,6 +837,25 @@ VkResult WINAPI __wine_create_vk_device_with_callback(VkPhysicalDevice phys_dev,
     res = wine_vk_device_convert_create_info(create_info, &create_info_host, &create_info_free_extensions);
     if (res != VK_SUCCESS)
         goto fail;
+
+    /* Enable shaderStorageImageWriteWithoutFormat for fshack
+     * This is available on all hardware and driver combinations we care about.
+     */
+    if (create_info_host.pEnabledFeatures)
+    {
+        features = *create_info_host.pEnabledFeatures;
+        features.shaderStorageImageWriteWithoutFormat = VK_TRUE;
+        create_info_host.pEnabledFeatures = &features;
+    }
+    if ((features2 = wine_vk_find_struct(&create_info_host, PHYSICAL_DEVICE_FEATURES_2)))
+    {
+        features2->features.shaderStorageImageWriteWithoutFormat = VK_TRUE;
+    }
+    else if (!create_info_host.pEnabledFeatures)
+    {
+        features.shaderStorageImageWriteWithoutFormat = VK_TRUE;
+        create_info_host.pEnabledFeatures = &features;
+    }
 
     if (native_vkCreateDevice)
         res = native_vkCreateDevice(phys_dev->phys_dev,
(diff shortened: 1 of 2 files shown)


dlls/winevulkan/vulkan.c
```diff
@@ -836,6 +838,25 @@ VkResult WINAPI __wine_create_vk_device_with_callback(VkPhysicalDevice phys_dev,
     if (res != VK_SUCCESS)
         goto fail;
 
+    /* Enable shaderStorageImageWriteWithoutFormat for fshack
+     * This is available on all hardware and driver combinations we care about.
```
ivyl, line 842: what combinations do we care about? what are the unsupported ones?

just sanity checking this statement

dlls/winevulkan/vulkan.c
```diff
@@ -1679,73 +1700,107 @@ void WINAPI wine_vkGetPrivateDataEXT(VkDevice device, VkObjectType object_type,
...
 {
     vec2 texcoord = (vec2(gl_GlobalInvocationID.xy) - constants.offset) / constants.extents;
     vec4 c = texture(texSampler, texcoord);
-    imageStore(outImage, ivec2(gl_GlobalInvocationID.xy), c.bgra);
+
+    // Convert linear -> srgb
+    bvec3 isLo = lessThanEqual(c.rgb, vec3(0.0031308f));
+    vec3 loPart = c.rgb * 12.92f;
+    vec3 hiPart = pow(c.rgb, vec3(5.0f / 12.0f)) * 1.055f - 0.055f;
+    c.rgb = mix(hiPart, loPart, isLo);
```
ivyl, line 1724: this looks okay to me but would be nice to also have someone more knowledgeable review the conversion :-)

dlls/winevulkan/vulkan.c
```diff
@@ -2889,269 +2841,24 @@ static VkResult record_compute_cmd(VkDevice device, struct VkSwapchainKHR_T *swa
...
-    blitregion.dstSubresource.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
-    blitregion.dstSubresource.layerCount = 1;
-    blitregion.dstOffsets[0].x = swapchain->blit_dst.offset.x;
-    blitregion.dstOffsets[0].y = swapchain->blit_dst.offset.y;
-    blitregion.dstOffsets[0].z = 0;
-    blitregion.dstOffsets[1].x = swapchain->blit_dst.offset.x + swapchain->blit_dst.extent.width;
-    blitregion.dstOffsets[1].y = swapchain->blit_dst.offset.y + swapchain->blit_dst.extent.height;
-    blitregion.dstOffsets[1].z = 1;
-
-    device->funcs.p_vkCmdBlitImage(hack->cmd,
```
ivyl, line 3117: > drop the vkCmdBlitImage path, it's impossible to keep without  VK_KHR_swapchain_mutable_format.

looks like ANV, RADV and Nvidia blobs support that extension

dlls/winevulkan/vulkan.c
DadSchoorse: > what are the unsupported ones?

Some mobile drivers and really old anv.

dlls/winevulkan/vulkan.c
DadSchoorse: Yes support is decent, but enabling and using extensions is kind of messy in winevulkan. There isn't really a good reason to keep the blit path, it might be slightly more efficient but fshack overhead is negligible in general.

dlls/winevulkan/vulkan.c
DadSchoorse: Actually, literally every driver+device on linux supports it: https://vulkan.gpuinfo.org/listdevicescoverage.php?feature=shaderStorageImageWriteWithoutFormat&platform=linux&option=not

The anv problem I remembered was probably format specific.

dlls/winevulkan/vulkan.c
doitsujin: The sRGB conversion code is taken from DXVK's HUD renderer which sometimes has to deal with sRGB-ness. This should be good.

dlls/winevulkan/vulkan.c
```diff
@@ -2244,7 +2203,7 @@ static VkResult init_fs_hack_images(VkDevice device, struct VkSwapchainKHR_T *sw
         viewInfo.sType = VK_STRUCTURE_TYPE_IMAGE_VIEW_CREATE_INFO;
         viewInfo.image = swapchain->fs_hack_images[i].user_image;
         viewInfo.viewType = VK_IMAGE_VIEW_TYPE_2D;
-        viewInfo.format = createinfo->imageFormat;
+        viewInfo.format = VK_FORMAT_B8G8R8A8_SRGB;
```
doitsujin, line 2206: What if the app requests something like RGB10A2 or RGBA16F as the swap chain format? Does fshack filter the supported swap chain formats anywhere?

DadSchoorse: No, but we have a big `FIXME("swapchain image format is not BGRA8 UNORM/SRGB. Things may go badly. %d\n", native_info.imageFormat);`. I doubt we will ever see anything but BGRA8 on X11.

ivyl:
Merged into experimental. Thanks!

PR #159, opened 2022-09-01
winevulkan: Check for NULL fence in vkDestoryFence.

DadSchoorse:


DadSchoorse:
guy already fixed this in bleeding edge

PR #160, opened 2022-09-09
windowscodecs: Cherry-pick upstream commit cdb9a274

pixelcluster:
This commit (https://gitlab.winehq.org/wine/wine/-/commit/cdb9a2748ba62891bfcd342cb3da82c204f896d2) contains a bugfix related to bitmap scaling. It fixes the `CopyPixels` function rejecting parameters that are valid on windows, where the destination bitmap buffer doesn't include padding for the last scanline. This behaviour is relied upon by DirectXTex's mipmap generation.

This commit fixes the black texture problem in BeamNG.drive reported in https://github.com/ValveSoftware/Proton/issues/1237#issuecomment-988444929.


ivyl:
This is now in bleeding-edge and if it doesn't cause any regressions should be a part of the next experimental release. Thanks!

https://github.com/ValveSoftware/wine/commits/experimental-wine-bleeding-edge-7.0-25275-20220919-p6c27a8-w2bc7af-d5c22e2-vc101da

PR #162, opened 2022-10-25
Backport "ntdll: Implement RtlRestoreContext on i386."

miniduikboot:
This commit was added to Wine 7.16: https://github.com/wine-mirror/wine/commit/88d13e663d1e6d797b4ea74a80a6e97e418e37f4

This commit fixes the BepInEx modloader on 32-bit IL2CPP Unity games like Among Us which would otherwise crash when a garbage collection cycle runs. Recent builds of BepInEx have switched from Mono to CoreCLR, which introduced this issue. BepInEx has built in a detection whether this patch is present: https://github.com/BepInEx/BepInEx/commit/a77efac44b8a94bf953b544b0508248894212d12#diff-dbb9378de52cf55fc473d0293ac660901f9b0b744d54081f9f23cc30f6e668e8R57

I've built Proton locally and confirmed that this commit builds and fixes the issue.


ivyl:
Upstream cherry-pick will be included in the next experimental release.

PR #163, opened 2022-10-26
winebus.sys: Enable hidraw for Simucube racing wheels and Fanatec pedals.

logos128:
Simucube FFB wheels have an additional function to control the maximum steering angle through vendor hid reports. Additionally, the devices are HID PID compliant and take full advantage of Wine 7's hidraw backend, which provides FFB through the HID PID implementation.
Fanatec ClubSport pedals have rumble motors on brake and throttle, which can be controlled via the Fanatec SDK (also using vendor hid reports).
Various apps/games use these functionalities and expect them to be available for these devices.

Fixes Raceroom Racing Experience crashing in hidapi.dll (provided by game) on Proton 7/Proton Experimental 7 when trying to load a racing session.

### DIFF
--- a/dlls/winebus.sys/bus_udev.c
+++ b/dlls/winebus.sys/bus_udev.c
@@ -1695,7 +1695,8 @@ static void udev_add_device(struct udev_device *dev, int fd)
         memcpy(desc.serialnumber, zeros, sizeof(zeros));
     }
 
-    if (!is_dualshock4_gamepad(desc.vid, desc.pid) && !is_dualsense_gamepad(desc.vid, desc.pid) && !is_thrustmaster_hotas(desc.vid, desc.pid))
+    if (!is_dualshock4_gamepad(desc.vid, desc.pid) && !is_dualsense_gamepad(desc.vid, desc.pid) && !is_thrustmaster_hotas(desc.vid, desc.pid) &&
+        !is_simucube_wheel(desc.vid, desc.pid) && !is_fanatec_pedals(desc.vid, desc.pid))
     {
         TRACE("hidraw %s: deferring %s to a different backend\n", debugstr_a(devnode), debugstr_device_desc(&desc));
         close(fd);
--- a/dlls/winebus.sys/unix_private.h
+++ b/dlls/winebus.sys/unix_private.h
@@ -286,5 +286,7 @@ BOOL is_dualshock4_gamepad(WORD vid, WORD pid) DECLSPEC_HIDDEN;
 BOOL is_dualsense_gamepad(WORD vid, WORD pid) DECLSPEC_HIDDEN;
 BOOL is_logitech_g920(WORD vid, WORD pid) DECLSPEC_HIDDEN;
 BOOL is_thrustmaster_hotas(WORD vid, WORD pid) DECLSPEC_HIDDEN;
+BOOL is_simucube_wheel(WORD vid, WORD pid) DECLSPEC_HIDDEN;
+BOOL is_fanatec_pedals(WORD vid, WORD pid) DECLSPEC_HIDDEN;
 
 #endif /* __WINEBUS_UNIX_PRIVATE_H */
--- a/dlls/winebus.sys/unixlib.c
+++ b/dlls/winebus.sys/unixlib.c
@@ -110,6 +110,24 @@ BOOL is_thrustmaster_hotas(WORD vid, WORD pid)
     return vid == 0x044F && (pid == 0xB679 || pid == 0xB687 || pid == 0xB10A);
 }
 
+BOOL is_simucube_wheel(WORD vid, WORD pid)
+{
+    if (vid != 0x16D0) return FALSE;
+    if (pid == 0x0D61) return TRUE; /* Simucube 2 Sport */
+    if (pid == 0x0D60) return TRUE; /* Simucube 2 Pro */
+    if (pid == 0x0D5F) return TRUE; /* Simucube 2 Ultimate */
+    if (pid == 0x0D5A) return TRUE; /* Simucube 1 */
+    return FALSE;
+}
+
+BOOL is_fanatec_pedals(WORD vid, WORD pid)
+{
+    if (vid != 0x0EB7) return FALSE;
+    if (pid == 0x183B) return TRUE; /* Fanatec ClubSport Pedals v3 */
+    if (pid == 0x1839) return TRUE; /* Fanatec ClubSport Pedals v1/v2 */
+    return FALSE;
+}
+
 struct mouse_device
 {
     struct unix_device unix_device;


srlemke:
Thanks for  also adding the simucube!
I have been trying those changes manually via regedit and simucube works flawleslly with wine >7, and have noted exact that problem with Raceroom, crash with Proton > 7 (while with Protoon 6 it works fine).

rbernon:
Thanks I've merged it, this will be in next Experimental release.

srlemke:
Hello, just a feedback, I have set to use the experimental -> bleeding edge(betas tab) proton, and can confirm that this makes the simucube work now, that experimental version also introduces DXVK 2.0 which fixes a lot of problems on raceroom, thanks all!

berniyh:
I can also confirm that FFB works now on the bleeding edge branch.
However, in both games I tried (Assetto Corsa Competizione and American Truck Simulator (via Proton, not native)), there is now something wrong with the joystick inputs overall.
If I assign the Simucube to the steering axis in ACC, it works, but says it's my shifter.
Assigning pedals says it's the Simucube, although they are a separate controller.
And finally one additional controller doesn't work (that's possibly the one that went over the cliff due to the rearrangement?).

Also, the pedals (being not centered by design) will behave as cursor keys, making the whole thing completely unusable, since it's skipping through the GUI buttons like crazy.

So regarding the controllers, from my observation, there is something seriously wrong with the bleeding edge branch.
But the FFB works, at least. :)

rbernon:
Hmm, sorry to hear that.

Would you mind attaching a log file, with normal Experimental (or Proton 7.0) where the mapping is correct, and another one in the bleeding edge branch where it isn't?

You can generate the log by setting `PROTON_LOG=1 WINEDEBUG=+pid,+tid,+dinput,+xinput,+input,+rawinput,+hid %command%` in the game start options, then running the game. After the game exits you should be able to find a file named `steam-<appid>.log` in your HOME directory. Mind copying it elsewhere before running it again with another Proton version as it gets overwritten.

Thanks!

logos128:
ACC does this if you reuse the same wine prefix created by another proton version. Since bleeding edge uses the hidraw backend for Simucube, while previous Proton versions used SDL, the controller sequence is probably different. One possible way to fix this is to remove the entire ACC prefix, and let Proton recreate a new one. This will wipe your ACC profile, so you'll have to start from scratch. Preferably you can just remove the controller setup file from your ACC proton prefix, which is probably located in <prefix>/drive_c/users/steamuser/Documents/Assetto Corsa Competizione/Config/controls.json , and reassign your controls again.

berniyh:
Deleting the prefix for ACC and starting with a clean prefix fixed the labels of the controllers, but that's it. The rest remains the same.
Here's a log for that:
https://murena.io/s/5twbxXFcj8J7TWd

I tried with various Proton versions and couldn't the controllers working.
I'm about 99% sure that I got it working at some point in the past (maybe half a year ago?) and that the FFB was all that was missing, but I don't play ACC that often, usually I'm using Assetto Corsa (244210).
For that, I know that switching Proton versions creates problems and as it's quite some work to get it set up, I did the test in ACC.
Nevertheless, in Assetto Corsa, all the controllers work fine (using Proton GE there, but I've also tested that one for ACC with a clean prefix). I've got a log for that one here:
https://murena.io/s/BFEoiPQkpPzwC93

The controllers I use are these ones:
```
Bus 007 Device 010: ID 16d0:0d5a MCS SimuCUBE
Bus 007 Device 009: ID 10c4:8b02 Silicon Labs HE SIM PEDALS
Bus 007 Device 008: ID 16c0:05b7 Van Ooijen Technische Informatica BU0836-LC Interface  <--- missing in ACC
Bus 003 Device 003: ID 044f:b660 ThrustMaster, Inc. T500 RS Gear Shift

```
Search for 16d0, 10c4 and 044f gives lines that they are added by the sdl backend for ACC, e.g.:
```
0060:0088:trace:hid:sdl_add_device` joystick id 0, desc {vid 044f, pid b660, version 0111, input -1, uid 00000000, is_gamepad 0}.
```
But I can't find a line mentioning 16c0, thus for some reason it's not added.

In the AC log, I don't find any lines mentioning 'sdl_add_device' and I also don't find any of those product (or vendor) IDs.
Thus I suspect it's the SDL backend that is at fault here and I suspect that this also the reason why there is now involuntary cursor movement when controllers are not centered (again, Pedals aren't, by design).
For ACC it could be that I it was working, because I might have created the original prefix years ago (thus not using SDL), but I'm not sure.
This might be related:
https://github.com/ValveSoftware/Proton/issues/4775

logos128:
> Search for 16d0, 10c4 and 044f gives lines that they are added by the sdl backend for ACC

Simucube is added via the hidraw backend. The SDL backend simply receives an add device event and fails (indicated on the next line of the log), due to the device already being added by the UDEV (hidraw) backend.

Seems the problems with the Heusinkveld pedals (vid 10c4) are caused by SDL recognizing them as a gamepad instead of a joystick. I had similar issues in the past with my pedals not being recognized by Proton games. Appeared, SDL refused to add them as a jpoystick, because it required at least one button declaration in addition to the axes.
You can try disabling Steam controller support in the settings, and see if that helps.
I'm pretty sure the pedals will work via the hidraw backend. They need to be enabled in a similar way as in this PR, though.

> But I can't find a line mentioning 16c0, thus for some reason it's not added.

Is this a button box? Seems SDL doesn't recognize it as either a joystick or a gamepad. Again going through the hidraw backend would probably fix the issue.

> In the AC log, I don't find any lines mentioning 'sdl_add_device' and I also don't find any of those product (or vendor) IDs.

You need to add the "WINEDEBUG=" part to the command line, otherwise it won't log the HID related channels.

rbernon:
Fwiw I just added support for a new `PROTON_ENABLE_HIDRAW` environment variable so that it's easier to test enabling the hidraw backend for individual devices. The feature should be available in the bleeding-edge branch of Proton Experimental if you want to try it (https://github.com/ValveSoftware/wine/commit/39539729e7c5a5171496f9b160152fa395203616).

You can set that variable to some string containing "`0xXXXX/0xYYYY`" (including the /, where 0xXXXX is the device VID, and 0xYYYY the PID each as 0x prefixed, 0-padded 4 char, hexadecimal numbers). Multiple devices should work too, as long as each has the same VID/PID format.

berniyh:
> Seems the problems with the Heusinkveld pedals (vid 10c4) are caused by SDL recognizing them as a gamepad instead of a joystick. I had similar issues in the past with my pedals not being recognized by Proton games. Appeared, SDL refused to add them as a jpoystick, because it required at least one button declaration in addition to the axes. You can try disabling Steam controller support in the settings, and see if that helps. I'm pretty sure the pedals will work via the hidraw backend. They need to be enabled in a similar way as in this PR, though.
> 
Interesting. I'm not sure though that going through hidraw here would be the best idea though.
The thing is that the pedals require (manual) calibration, which I did via hwdb.
If the hidraw backend is used, this calibration is lost within proton and has to be done for each proton prefix again.
Since wine (or proton) to my knowledge do not provide ways to do the calibration, I would need to run something like dxview in each prefix.
So for such a device, in the long term, it might be a better idea to keep using the SDL backend and fix the issues with it.

> > But I can't find a line mentioning 16c0, thus for some reason it's not added.
> 
> Is this a button box? Seems SDL doesn't recognize it as either a joystick or a gamepad. Again going through the hidraw backend would probably fix the issue.
> 
It's a handbrake. Of course the handbrake in ACC is more or less useless, but my two push/pull shifters are connected to it, so I would need it in ACC as a shifter.
Also, it would suffer from the same problem regarding the calbration.

> > In the AC log, I don't find any lines mentioning 'sdl_add_device' and I also don't find any of those product (or vendor) IDs.
> 
> You need to add the "WINEDEBUG=" part to the command line, otherwise it won't log the HID related channels.
>
A right, I totally forgot. Sorry, my mistake.
Nevertheless, I guess it's using a different backend. I don't see any other explanation.

> Fwiw I just added support for a new `PROTON_ENABLE_HIDRAW` environment variable so that it's easier to test enabling the hidraw backend for individual devices. The feature should be available in the bleeding-edge branch of Proton Experimental if you want to try it ([3953972](https://github.com/ValveSoftware/wine/commit/39539729e7c5a5171496f9b160152fa395203616)).
> 
> You can set that variable to some string containing "`0xXXXX/0xYYYY`" (including the /, where 0xXXXX is the device VID, and 0xYYYY the PID each as 0x prefixed, 0-padded 4 char, hexadecimal numbers). Multiple devices should work too, as long as each has the same VID/PID format.

Thanks, I'll give that a go. Multiple devices would be separated by a comma?

berniyh:
Here's the log including the debug info for Assetto Corsa, which works for all controllers (apart from the FFB for the Simucube, since I ran it with Proton GE, not bleeding edge):
[steam-244210.zip](https://github.com/ValveSoftware/wine/files/10146798/steam-244210.zip)

Couldn't get the new env var working yet, but maybe my proton is still too old, will try again tomorrow.

logos128:
> Interesting. I'm not sure though that going through hidraw here would be the best idea though.
The thing is that the pedals require (manual) calibration, which I did via hwdb.
If the hidraw backend is used, this calibration is lost within proton and has to be done for each proton prefix again.

IMO, you can do it with Heusinkveld's app in a similar way as Simucube's TrueDrive (in a dedicated wineprefix outside of proton, etc.). To my knowledge it saves the calibration directly to the pedals board.

logos128:
@rbernon
Thanks. This will make things much easier.

berniyh:
> > Interesting. I'm not sure though that going through hidraw here would be the best idea though.
> > The thing is that the pedals require (manual) calibration, which I did via hwdb.
> > If the hidraw backend is used, this calibration is lost within proton and has to be done for each proton prefix again.
> 
> IMO, you can do it with Heusinkveld's app in a similar way as Simucube's TrueDrive (in a dedicated wineprefix outside of proton, etc.). To my knowledge it saves the calibration directly to the pedals board.

Only for the newer models that feature another controller.
tbh, in a way, I'm glad that I don't have it, because I can calibrated them very easily using udev/hwdb.

berniyh:
Also, I grabbed this test tool, to check SDL:
https://github.com/meleu/jstest-sdl

jstest-sdl -l gives me:
```
Found 5 joystick(s)

Joystick Name:     'Thustmaster T500 RS Gear Shift'
Joystick GUID:     030053514f04000060b6000011010000
Joystick Number:    0
Number of Axes:     2
Number of Buttons: 10
Number of Hats:     0
Number of Balls:    0
GameController:
  not a gamepad

Joystick Name:     'Heusinkveld Engineering HE SIM PEDALS'
Joystick GUID:     03007ff1c4100000028b000011010000
Joystick Number:    1
Number of Axes:     7
Number of Buttons: 12
Number of Hats:     0
Number of Balls:    0
GameController:
  Name:    'Heusinkveld Engineering HE SIM PEDALS'
  Mapping: '03007ff1c4100000028b000011010000,Heusinkveld Engineering HE SIM PEDALS,a:b0,b:b1,x:b3,y:b4,back:b10,start:b11,leftshoulder:b6,rightshoulder:b7,leftx:a0,lefty:a1,rightx:a3,righty:a4,lefttrigger:a2,righttrigger:a5,crc:f17f,platform:Linux'

Joystick Name:     'Granite Devices SimuCUBE'
Joystick GUID:     030037efd01600005a0d000011010000
Joystick Number:    2
Number of Axes:     8
Number of Buttons: 79
Number of Hats:     0
Number of Balls:    0
GameController:
  not a gamepad

Joystick Name:     'Simple Solutions ButtonBox'
Joystick GUID:     0300bacc9016000013fe000010010000
Joystick Number:    3
Number of Axes:     4
Number of Buttons: 32
Number of Hats:     0
Number of Balls:    0
GameController:
  not a gamepad

Joystick Name:     'Leo Bodnar BU0836-LC Interface'
Joystick GUID:     0300424fc0160000b705000010010000
Joystick Number:    4
Number of Axes:     1
Number of Buttons: 32
Number of Hats:     1
Number of Balls:    0
GameController:
  not a gamepad
```
So yeah, the HE SIM Pedals are recognized as a game controller by SDL for some reason. The lack of buttons can't be the reason, since it clearly lists them. Maybe it's the number of axes? 
It also recognizes the BU0836-LC interface (which is the handbrake that is missing in proton).

berniyh:
Ok, so I got it working (indeed, a comma-separated list) and that fixes the problems, but it brings up new ones.
As I correctly assumed, I'd have to do the calibration inside the prefix using DIView and that does not work.
The controllers are recognized and I can view both the raw and the output values, but I think wine does something wrong here.
The controller of both HE devices (The HE SIM PEDALS and the BU0836-LC interface) are 12 Bit controllers, so 4096 possible values and I correctly states 4096 as the maximum on the raw axis.
The actual values (when moving the controller) seem to be 16 Bit (like the output value) and thus go out of range.
Also, I can't set the calibration, when I try to do so, I get an error message "ERROR: SetProperty returned 0x80004001" and in the console output:
```
0280:fixme:dinput:dinput_device_set_property Unknown property <guid-0x000b>
```
image


Of course, this is something that should be fixed, but overall, I think the proper solution here is to fix the SDL backend for these devices. Going hidraw for everything might not be the best idea.
Nevertheless, I was able to work around it by adjusting the gamepad mapping for the HE SIM Pedals such that the x and y axis are assigned to rightx and righty instead of leftx and lefty. Then they don't move around the menu. Not the best solution, but it works.
Had to go through steam's config.vdf though, since for some reason they are not listed in steam's controller menu.
And then use hidraw for the BU0836-LC (which however only works for games in which I don't need the handbrake axis, since I can't calibrate that axis properly).

The whole thing is very weird, tbh.

Edit: forgot to say: Steam Input is disabled for all of these tests in ACC's steam properties. I tried with enabled, didn't see a difference.

PR #165, opened 2022-11-24
winevulkan: Update to 1.3.235 spec and enable 64bit VK_EXT_descriptor…

DadSchoorse:
…_buffer.

32bit needs better struct conversion.


ivyl:
Merged into experimental. Should appear in bleeding edge soon.

PR #166, opened 2022-11-26
winevulkan: Use direct calls for vkGetDescriptorEXT.

DadSchoorse:
As performance sensitive as descriptor updates.


ivyl:
this has been merged into experimental, thanks!

PR #169, opened 2022-12-21
winex11.drv: Revert workaround for mutter issue #676, fixing issue #2438

UrbanSombrero:
The workaround for mutter issue #676 prevents unredirected fullscreen windows from updating their decorations on mutter, breaking minimize for these windows on mutter 43 (mutter issue #2438). Although #676 remains open, the bug was fixed several years ago, and the test program attached to that issue behaves correctly on mutter 3.30 and up. As such, this hack can be safely removed, which will resolve issue #2438.

Additionally, mutter issue #306 has been fixed since mutter 3.28 and the workaround for it should also now be removed, as per its comment.

https://gitlab.gnome.org/GNOME/mutter/-/issues/2438

### DIFF
--- a/dlls/winex11.drv/event.c
+++ b/dlls/winex11.drv/event.c
@@ -1525,19 +1525,6 @@ static BOOL X11DRV_PropertyNotify( HWND hwnd, XEvent *xev )
         XFree(name);
     }
 
-    if (event->atom == x11drv_atom(_NET_WM_BYPASS_COMPOSITOR))
-    {
-        struct x11drv_win_data *data = get_win_data( hwnd );
-        if (!data) return TRUE;
-
-        /* workaround for mutter gitlab bug #676, changing decorations of a
-         * fullscreen and unredirected window freezes the compositing.
-         */
-        if (wm_is_mutter( data->display )) set_wm_hints( data );
-
-        release_win_data( data );
-    }
-
     if (event->atom == x11drv_atom(WM_STATE)) handle_wm_state_notify( hwnd, event, TRUE );
     else if (event->atom == x11drv_atom(_NET_WM_STATE)) handle__net_wm_state_notify( hwnd, event );
     return TRUE;
--- a/dlls/winex11.drv/window.c
+++ b/dlls/winex11.drv/window.c
@@ -874,33 +874,6 @@ static void set_mwm_hints( struct x11drv_win_data *data, DWORD style, DWORD ex_s
     MwmHints mwm_hints;
     int enable_mutter_workaround, mapped;
 
-    /* workaround for mutter gitlab bug #676, changing decorations of a
-     * fullscreen and unredirected window freezes the compositing.
-     * The window style will be updated again once the window has returned
-     * from fullscreen.
-     */
-    if (wm_is_mutter(data->display) && (data->net_wm_state & (1 << NET_WM_STATE_FULLSCREEN)))
-    {
-        Atom type;
-        int format;
-        unsigned long *property, net_wm_bypass_compositor = 0, count, remaining;
-
-        if (XGetWindowProperty( data->display, data->whole_window, x11drv_atom(_NET_WM_BYPASS_COMPOSITOR), 0,
-                                1, False, XA_CARDINAL, &type, &format, &count, &remaining,
-                                (unsigned char **)&property ) == Success &&
-            property)
-        {
-            net_wm_bypass_compositor = *property;
-            XFree(property);
-        }
-
-        if (net_wm_bypass_compositor)
-        {
-            TRACE("workaround mutter bug, ignoring decorations while compositor is bypassed\n");
-            return;
-        }
-    }
-
     if (data->hwnd == GetDesktopWindow())
     {
         if (is_desktop_fullscreen()) mwm_hints.decorations = 0;


rbernon:
Sadly we recently found that although the initial issues that required mutter #306 workaround, and the similar one for KDE, seem to be fixed, the workaround is still useful in some cases. More precisely, they allow to alt-tab out of fullscreen games, which would otherwise not be possible.

Specifically there are two cases where windows may be fullscreen and top-most. First, applications can do that directly, changing the windows decoration, size and top-most attribute. Additionally, on Windows there is an "Fullscreen Exclusive" mechanism, which D3D uses and which doesn't exist on Linux. Wine emulates this mechanism by changing the decorations, size and top-most attributes of the windows in an indistinguishable way.

On Windows, such "Fullscreen Exclusive" mode is automatically disabled when the window loses focus, but that's not the case if the window was manually set top-most by an application. We currently have no way to tell between the two cases, and we cannot automatically change the windows properties when the window loses focus. This means that a window that has been set top-most and fullscreen will stay over other windows even when it loses focus, unless we skip the top-most attribute.

(Note that it is also a very brittle area, full of race conditions, and any change is very likely to break something else)

UrbanSombrero:
That may have been the case in the past, but mutter has been correctly handling alt-tab switching in fullscreen Wine games for several years now. (As mentioned above, the mutter-side fix for #306 has been backported as early as 3.28). The reason *Proton* still requires the #306 hack is because of the #676 hack, which prevents decoration changes (and is currently responsible for breaking alt-tab switching in mutter 43). With windows being permitted to change decorations when losing focus, the top-most attribute is no longer a problem.

I understand that there's a delicate arrangement of dominoes involved here; I've had to do extensive testing to address this issue. This has been a source of frustration on GNOME for months now, and I hope you'll give this a closer look. If you'd rather play it safe and only revert #676, that would be enough to fix the problem, but I think that removing both is reasonable.

qwertychouskie:
Any progress on this issue?  As more distros upgrade to Gnome 43, this issue is only going to become more wide-spread.

LethalManBoob:
> Any progress on this issue? As more distros upgrade to Gnome 43, this issue is only going to become more wide-spread.

+1 this is a major hindrance to my ux.

qwertychouskie:
@UrbanSombrero It look like there's some sort of conflict/failed rebase/merge/etc.

UrbanSombrero:
Right, just fixed it.

UrbanSombrero:
You're right, I went back and tested older versions of GNOME and the #306 workaround does in fact break alt-tab switching on mutter < 43, so I'm removing that commit from this PR. Removing #306 worked for me on mutter 43 and vanilla Wine works on older versions with the top-most attribute set, but clearly Proton behaves differently, my mistake.

More importantly, though, the #676 workaround is currently breaking alt-tab switching from fullscreen windows on GNOME 43, because it depends on mutter allowing minimizing without the NET_WM_ACTION_MINIMIZE hint, which is no longer the case. I've now tested older versions of GNOME to verify that Proton fullscreen windows on 3.30 through 42 are unaffected by this proposed change.

UrbanSombrero:
For anyone who wants to try this out for themselves, I wrote a GitHub Actions workflow to automate building Proton from the experimental_7.0 branch with this PR applied, and you can download it from here:
https://github.com/UrbanSombrero/Proton/releases

Extract to `~/.steam/root/compatibilitytools.d`, restart Steam, and select it as your Steam Play compatibility tool in the settings.

(To help reduce comment noise, just leave an emoji reaction if it works for you.)

ivyl:
> More importantly, though, the #676 workaround is currently breaking alt-tab switching from fullscreen windows on GNOME 43, because it depends on mutter allowing minimizing without the NET_WM_ACTION_MINIMIZE hint, which is no longer the case. I've now tested older versions of GNOME to verify that Proton fullscreen windows on 3.30 through 42 are unaffected by this proposed change.

3.30 was released in 2018. The bug report is from 2019 and explicitly states that `(Tested with mutter 3.32.2)`. I'm pretty certain that I've also reproduce it in late 2020.

Are you sure you've use the test program correctly?

UrbanSombrero:
I'm fairly certain, I was able to reproduce the issue on 3.28 but 3.30 worked fine for me.

UrbanSombrero:
I just went back and double-checked, Fedora 29 (3.30) worked properly, as did every later version I tried, while Fedora 28 (3.28) behaved exactly as described in the issue.

ivyl:
I was able to reproduce this with Fedora 30 which comes with `mutter-3.32.1-1.fc30.x86_64`. I've downloaded the updated example from [the comment](https://gitlab.gnome.org/GNOME/mutter/-/issues/676#note_563983). Just a few random presses of `f`, `u` and `d` is enough for mutter to stop refreshing. This is reliable for me - 3 out of 3 times on frame < 20.

LethalManBoob:
> Author

Looks to have fixed my alt tabbing issue. Will report back if somthing's still broken.

ivyl:
Looks like it indeed may have got fixed somewhere between 3.32 and 3.34. I'll have to have a look at the games that are known to have required the work around in the first place to be sure they won't regress. If everything will be fine this should land in experimental in the next few days.

LeJimster:
I tested the build @UrbanSombrero posted and it does indeed fix the alt+tab issue.   I can now tab out of the game and go back no issue.   I'm using Trackmania United Forever as an example.
The only minor issue is the bug still appears visible in the preview windows when alt+tabbing or on the activities screen (super key).
Screenshot from 2023-01-04 13-38-50

ivyl:
A cleaned up revert was merged into bleeding-edge and if everything goes well it should be released in an experimental this week.

ivyl:
Alt+Tab should now work with Gnome 43 on the latest experimental. See: https://github.com/ValveSoftware/Proton/wiki/Changelog

PR #170, opened 2022-12-23
fixup! dinput: Avoid sending envelope report if it has been removed. 

logos128:
Prevent the flags variable from re-raising DIEP_ENVELOPE in impl->flags.

@rbernon , this is based on experimental-wine-bleeding-edge-7.0-31663-20221221 .


rbernon:
Thanks! I've merged the fixup into experimental, it should be in bleeding edge branch soon. I don't think there's any plan to have an experimental release already, as we just did one, but it will be in the next one.

PR #171, opened 2022-12-29
Revert "HACK: proton: ntdll,server: Never use CLOCK_MONOTONIC_RAW"

ishitatsuyuki:
This reverts commit aba1a41c5c5d430b40466d4f347aad47c5087663.

Per [1], the performance disparity has been fixed in Linux 5.3.

The hack also made vkGetCalibratedTimestampsEXT calibrate to the wrong clock, since the code to choose the time domain was designed to mirror upstream Wine behavior. A possible fix is to mirror the hack over to vkGetCalibratedTimestampsEXT, however since Linux 5.3 has been around for a while, removing the hack seems to be the better option.

[1] https://lore.kernel.org/linux-arm-kernel/20190621095252.32307-1-vincenzo.frascino@arm.com/

### DIFF
--- a/dlls/ntdll/unix/sync.c
+++ b/dlls/ntdll/unix/sync.c
@@ -92,7 +92,7 @@ static inline ULONGLONG monotonic_counter(void)
     return mach_absolute_time() * timebase.numer / timebase.denom / 100;
 #elif defined(HAVE_CLOCK_GETTIME)
     struct timespec ts;
-#if 0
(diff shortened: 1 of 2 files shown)


ivyl:
Thanks! Makes a lot of sense to revert this now. 5.3 was a long time ago already. I've merged this into the experimental tree, it should show up in bleeding-edge soon.

Issue #177, opened 2023-02-19
DxDiag returning incorrect values


LeonTheo02:
When querying DxDiag on the Steam Deck it will return the version "DirectX 9.0c". If a game or software uses this to check if the required DirectX version is availible it will fail the check and not allow it to launch when a higher DirectX version that the hardware supports is required. Should this not be increased to return the correct values for DirectX 12?

PR #178, opened 2023-02-23
DxDiag returning DirectX 12 values

LeonTheo02:
Fixing #177 

Changed the values DxDiag returns to reflect values taken from a modern DirectX 12 capable system via Windows DxDiag. This will allow games that check these values to detect the DirectX Version correctly.


ivyl:
Do you know of any games that actually do check this? We haven't stumbled upon anything that would use it.

LeonTheo02:
We do on RAID: World War II for example (the DirectX 11 Open Beta is currenlty not playable on Deck because of this) and its probably best to make DxDiag return what the Deck is capable of anyway.

ivyl:
Thanks! It's always helpful to know what real software is impacted by a given issue. I've merged it to experimental - it's already in bleeding-edge and should show up in our next experimental-proper release.

https://github.com/ValveSoftware/wine/commits/experimental-wine-bleeding-edge-7.0-36999-20230306-pffd4f0-wda59f6-d7f21a6-vbd22ca

PR #179, opened 2023-02-23
Override Bloodrayne Terminal Cut 1/2 to use built in DX8-9 wrapper.

Bitwolfies:
Proton doesn't use the games wrapper by default, causing WineD3D to be used over DXVK, which causes massive slowdown and a few misc bugs.


ivyl:
Hi. I've pushed this to experimental and it's already part of bleeding-edge.

https://github.com/ValveSoftware/wine/commits/experimental-wine-bleeding-edge-7.0-37012-20230306-pffd4f0-w2e259b-d7f21a6-vbd22ca

I've changed the wording a bit because built-in has a very specific meaning in Wine. I instead refer to the d3d8.dll that ships with the game as the "included" one.

PR #184, opened 2023-03-18
Patches to for getting Revive working

twhitehead:
This patch adds the Windows fallback behaviour to calling [CertGetNameStringW](https://learn.microsoft.com/fr-fr/windows/win32/api/wincrypt/nf-wincrypt-certgetnamestringw) with `dwType=CERT_NAME_ATTR_TYPE` and `pvTypePara=NULL` (Windows returns the first existing name entry for email, CN, OU, or O, while Wine only tried to return email).

It is required to get the LibOVR runtime dll to verify, which is required in order to get Revive to load

https://github.com/LibreVR/Revive/issues/1571#issuecomment-1465240925

Hopefully this will now mean Oculus VR program can be ran under Revive, but, my understanding is, Wine lacks OpenXR/OpenVR support, so this cannot be tested without a Proton release with these patches.

They have also been submitted upstream to Wine too. Unfortunately they missed the 8.4 release by a matter of hours

https://gitlab.winehq.org/wine/wine/-/merge_requests/2403


twhitehead:
Just a note that this has [now been merged](https://gitlab.winehq.org/wine/wine/-/merge_requests/2403#note_1148d31156c23c2f2f262c482a91d0db8054000a) into wine.

rbernon:
Thanks, I've merge this into Experimental, it should be available soon in the bleeding-edge builds and in the next Experimental release (unless we find something wrong with it in the meantime).

twhitehead:
Sounds good. Thank's very much! :+1:

PR #192, opened 2023-06-22
wine.inf: Set msvcp140_atomic_wait to native,builtin.

Etaash-mathamsetty:
Regression from proton 6.3 (which didn't have msvcp140_atomic_wait). ~~Without this patch, there is a mismatch between native msvcp and built-in msvcp causing an unimplemented function crash/exception.~~ 
some games can call unimplemented methods in the built-in version of the dll, so set it to native (like all the others)


ivyl:
The change help with SDL2 beta of Dwarf Fortress that otherwise crashes on unimplemented `msvcp140_atomic_wait.dll.__std_acquire_shared_mutex_for_instance` in our built-in.

I've merged it. Everything should show up in the bleeding-edge soon.

PR #202, opened 2023-10-16
gdi32: Add stub for D3DKMTQueryAdapterInfo.

Etaash-mathamsetty:
Fixes rainbow 6 siege from not launching due to this missing function. backport of https://gitlab.winehq.org/wine/wine/-/merge_requests/3777. games that use DLSS3 will also call this function


GloriousEggroll:
Should be noted the game still also needs this envvar set from the mesa side:
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25734

We've had an envvar set for it in proton-ge for a long time

PR #210, opened 2023-12-14
ntdll: Force gstreamer to only use x11 for GST_GL_WINDOW

GloriousEggroll:
Proton only uses x11, however users using it outside of the steamos/steamdeck environment (such as on wayland desktops or with other applications such as Lutris, Heroic, Bottles, and so on) tend to hit a bug with gstreamer:

"GStreamer with OpenGL creates an empty "OpenGL Renderer" window on Wayland" https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/2997

There is an upstream merged MR in gstreamer that adds support for EGL_MESA_platform_surfaceless:

https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/5511

However this is not necessary if we instead force GST_GL_WINDOW=x11 even on wayland, it avoids the empty "OpenGL Renderer" windows on wayland.


ivyl:
It would make more sense to have it as a fixup for the downstream commit that enables OpenGL and have the `setenv()` somewhere in winegstreamer itself.

rbernon:
Yeah, fwiw the relevant commit to fixup is https://github.com/ValveSoftware/wine/commit/a4eb2896193314ad2d6123660c7a5e1da7e60e08. Probably doing `setenv` right before the `gst_gl_display_new` call should be enough?

GloriousEggroll:
> Yeah, fwiw the relevant commit to fixup is [a4eb289](https://github.com/ValveSoftware/wine/commit/a4eb2896193314ad2d6123660c7a5e1da7e60e08). Probably doing `setenv` right before the `gst_gl_display_new` call should be enough?

Updated

GloriousEggroll:
closing -- this breaks gstreamer when running inside proton + proton container. it only works when using proton's wine build outside the proton environment

GloriousEggroll:
-EDIT-

This is still a problem but the patch here doesn't resolve it. I'm still working on it.

PR #217, opened 2024-01-20
winex11: fix missing null check in glxdrv_wglSwapBuffers.

Etaash-mathamsetty:
fixes an exception in Black desert online's launcher. launcher is still black tho :(

### DIFF
--- a/dlls/winex11.drv/opengl.c
+++ b/dlls/winex11.drv/opengl.c
@@ -4926,7 +4926,7 @@ static BOOL glxdrv_wglSwapBuffers( HDC hdc )
     update_window_surface( gl, hwnd );
     release_gl_drawable( gl );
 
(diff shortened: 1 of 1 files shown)


gofman:
This is best to send upstream, no reason to have this as Proton specific especially since it is not known to really fix anything.

Maybe on the not entirely related note, it is interesting what is the actual problem there as NULL ctx may indicate the prior problem. Anyway, the patch might be valid as is but the place of it and the related discussion is Wine upstream.

Etaash-mathamsetty:
> This is best to send upstream, no reason to have this as Proton specific especially since it is not known to really fix anything.
> 
> Maybe on the not entirely related note, it is interesting what is the actual problem there as NULL ctx may indicate the prior problem. Anyway, the patch might be valid as is but the place of it and the related discussion is Wine upstream.

yep, i didn't check upstream code, same issue is there, will send upstream

PR #219, opened 2024-02-17
ntdll: Don't leak objattr allocation in NtCreateSemaphore.

ishitatsuyuki:
fsync/esync_create_semaphore allocates objattr on their own, and objattr allocated in NtCreateSemaphore was getting leaked.

Like NtCreateEvent, move the objattr allocation after fsync/esync early return to avoid double allocation.


ishitatsuyuki:
@gofman or @zfigura, would you mind taking a look when you have time?

gofman:
Thanks! Pushed to Experimental branch.

PR #222, opened 2024-03-07
winex11.drv: Apply the GL vendor override for Intel GPUs too

renatopereyra:
The vendor override forces wine to report an AMD GPU. Previously,
the vendor override only applied for NVIDIA GPUs.
    
Also, rename WINE_GL_HIDE_NVIDIA to WINE_GL_VENDOR_REPORT_AMD.

More info: https://github.com/ValveSoftware/Proton/issues/7529

### DIFF
--- a/dlls/winex11.drv/opengl.c
+++ b/dlls/winex11.drv/opengl.c
@@ -3494,7 +3494,7 @@ static const GLubyte *wglGetString(GLenum name)
         int sz;
 
         override_vendor = 0;
-        if ((env = getenv("WINE_GL_HIDE_NVIDIA")))
+        if ((env = getenv("WINE_GL_VENDOR_REPORT_AMD")))
         {
             override_vendor = env[0] != '0';
         }
@@ -3522,13 +3522,19 @@ static const GLubyte *wglGetString(GLenum name)
         if (name == GL_RENDERER)
         {
             s = pglGetString(name);
-            if (s && strstr((const char *)s, "NVIDIA")) return (const GLubyte *)"AMD Radeon Graphics";
+            if (s && (strstr((const char *)s, "NVIDIA") || strstr((const char *)s, "Intel")))
+            {
+                return (const GLubyte *)"AMD Radeon Graphics";
+            }
             return s;
         }
         else if (name == GL_VENDOR)
         {
             s = pglGetString(name);
-            if (s && strstr((const char *)s, "NVIDIA")) return (const GLubyte *)"AMD";
+            if (s && (strstr((const char *)s, "NVIDIA") || strstr((const char *)s, "Intel")))
+            {
+                return (const GLubyte *)"AMD";
+            }
             return s;
         }
     }


renatopereyra:
PTAL @gofman 

I'm not sure whether this repo strives to maintain backward compatibility for some number of releases. If so, I can add support for `WINE_GL_HIDE_NVIDIA` back and print some deprecation message.

gofman:
Thanks for the patch! I don't think we need to keep old env var name anyhow. This is not auto set outside of this place and such options is not a part of supported end user usage, we don't keep backward compat on hack options. I have a few comments for the rest though:

```
+    static char fake_renderer[64] = "AMD Radeon Graphics";
+    static char fake_vendor[32] = "AMD";
```
Let's call it amd_renderer, amd_vendor. And make it static const char in the view of the below comment. Or, rather just remove it and keep inline constant (see below).

```
+            if ((env = getenv("WINE_GL_VENDOR")))
+            {
+                override_vendor = 1;
+                lstrcpynA(fake_vendor, env, sizeof(fake_vendor));
+            }
+            if ((env = getenv("WINE_GL_RENDERER")))
+            {
+                override_vendor = 1;
+                lstrcpynA(fake_renderer, env, sizeof(fake_renderer));
+            }
```
Let's not do this, we don't need extra options added just in case.

```
+                override_vendor = 1;
+                lstrcpynA(fake_renderer, env, sizeof(fake_renderer));
+            }
         }
+
     }
```
Spurious newline.

```
        const GLubyte *s;
         if (name == GL_RENDERER)
         {
-            s = pglGetString(name);
-            if (s && strstr((const char *)s, "NVIDIA")) return (const GLubyte *)"AMD Radeon Graphics";
-            return s;
+            return (const GLubyte *)fake_renderer;
         }
         else if (name == GL_VENDOR)
         {
-            s = pglGetString(name);
-            if (s && strstr((const char *)s, "NVIDIA")) return (const GLubyte *)"AMD";
-            return s;
+            return (const GLubyte *)fake_vendor;
         }
```

We shouldn't fake anything if the vendor is already AMD, it is better to keep the original returned strings unchanged. So in general I suggest make much less changes, just rename the variable, possibly not declare amd_renderer, amd_vendor (keep them as inline constants) and check explicitly for the vendor / renderer we want to override. I. e., just ```strstr((const char *)s, "NVIDIA") -> ( (strstr((const char *)s, "NVIDIA") || strstr((const char *)s, "<proper subtsring for Intel>"))``` .

In general, options like that are a pain rather than a benefit, it is better to keep them as short as possible and focused rather making them gorgeous and multivariant.

renatopereyra:
> Let's not do this, we don't need extra options added just in case.

The thinking was that this makes sure we can always do this fully in Proton (perhaps from the python script?), which hopefully would remove the incentive to do this type of thing in Mesa. I'm not very familiar with this codebase and understand there may be reasons I am not aware of so I will remove it if you feel strongly. Let me know

> We shouldn't fake anything if the vendor is already AMD

Good catch, will do. Thanks!

gofman:
> > Let's not do this, we don't need extra options added just in case.
> 
> The thinking was that this makes sure we can always do this fully in Proton (perhaps from the python script?), which hopefully would remove the incentive to do this type of thing in Mesa. I'm not very familiar with this codebase and understand there may be reasons I am not aware of so I will remove it if you feel strongly. Let me know
> 
We can do it once needed and once there is a need to add something like that to proton script, we don't need an extra hack code just in case.

gofman:
Sorry, could you please still explicitly check for what we need to override and default to not overriding (like that hack was doing before and as I suggested in the comments above), just adding Intel) instead of overriding by default? This is an unfortunate hack and we want to be explicit about which cases we override. Also, Mesa was changing GL strings in the past, we want normal AMD path which works without hacking to be safe.

renatopereyra:
I'm sorry, I didn't realize that would be a point of concern and I was trying to keep the source succinct. Fixed.

gofman:
Thanks! It is in Proton Experimental [bleeding-edge] branch now (https://github.com/ValveSoftware/wine/commits/experimental-wine-bleeding-edge-9.0-84776-20240308-p9ca19d-w9f1b44-d69a52b-vffb04f/), it should appear in next Experimental proper release and then eventually settle in Proton Stable.

PR #231, opened 2024-04-16
wg_parser: add wow64 func for wg_parser_connect

peterjc123:


kisak-valve:
As noted at https://github.com/ValveSoftware/wine/issues/232#issuecomment-2058097104, it would make more sense for this to go into upstream wine.

peterjc123:
@kisak-valve It seems that this patch only applies here, not for upstream wine.

PR #235, opened 2024-05-02
winegstreamer: Fix wow64 support for wg_parser_connect.

peterjc123:
Wine-Bug: https://bugs.winehq.org/show_bug.cgi?id=56595

Commit in upstream wine: https://gitlab.winehq.org/wine/wine/-/commit/87eaf899551077817106bce62f5a8b980a47635f

Backport to fix games that uses gstreamer in wow64 mode.


peterjc123:
@Plagman Rebased.

PR #236, opened 2024-05-04
winebus.sys: Enable hidraw for VKB Gunfighter SCG

her001:
I have VKB Gunfighter joysticks with the Space Combat Grips which SDL is handling as Xbox controllers. Adding the PIDs for the left and right grips should fix this.


rbernon:
Merged, should be in bleeding-edge soon, and in next Proton Experimental (and later, stable) release.

PR #237, opened 2024-05-09
Backport std_smf_hypot3

Dandraghas:
Upstream wine has implemented `__std_smf_hypot3` and `__std_smf_hypot3f` in msvcp140_2 which fixes crashes in some games that rely on this.


kisak-valve:
Hello @Dandraghas, please note the games you know are affected so that it's easier for a Proton dev to have a test case.

Issue #238, opened 2024-05-20
Compilation error "ntsyscalls.h"


daviDarthemis:
dlls/ntdll/signal_arm64.c:36: error: ntsyscalls.h: No such file or directory

ivyl:
see: https://github.com/ValveSoftware/wine/pull/239#issuecomment-2157810486

Issue #221, opened 2024-02-25
ntsyscalls.h missing?


Goldsrc233:
when configuring:
dlls/ntdll/signal_arm.c:34: error: ntsyscalls.h: No such file or directory

lidaxian121:
May I ask how you solved this problem?

ivyl:
./autogen.sh

PR #239, opened 2024-05-31
Add call tools/make_specfiles to autogen.sh

jhollingworth:
If you don't call tools/make_specfiles then build will fail saying it's missing ntsyscalls.h. 

This confused me and a few others (https://github.com/ValveSoftware/wine/issues/221, https://github.com/ValveSoftware/wine/issues/238). Adding the call to autogen.sh to avoid this issue in the future

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 set -e
 tools/make_requests
+tools/make_specfiles
 dlls/winevulkan/make_vulkan -x vk.xml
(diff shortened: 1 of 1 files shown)


ivyl:
I've moved the `make_specfiles` after `make_vulkan` as it touches one of those files. Shouldn't matter much though.

I've also changed the commit message a bit to adhere to the style used by Wine. This should appear shortly in bleeding-edge and then will trickle to all the other branches.

Thanks!

PR #241, opened 2024-06-16
Update fa.po

aminkasani:
updated some of the translations


rbernon:
Thanks! Could you please create a merge request with these for the Wine project (https://gitlab.winehq.org/wine/wine)? It will benefit a larger audience and we can then cherry-pick the changes in Proton.

PR #245, opened 2024-07-05
mmdevapi: Support getting IAudioClockAdjustment interface.

Etaash-mathamsetty:
Allows ghostbusters the video game remastered (ID 1449280) to play the 4K high definition videos through mfplat. since mfplat is finally able to handle this game :)

### DIFF
--- a/dlls/mmdevapi/client.c
+++ b/dlls/mmdevapi/client.c
@@ -880,6 +880,9 @@ static HRESULT WINAPI client_GetService(IAudioClient3 *iface, REFIID riid, void 
     } else if (IsEqualIID(riid, &IID_IAudioClock)) {
         IAudioClock_AddRef(&This->IAudioClock_iface);
         *ppv = &This->IAudioClock_iface;
+    } else if (IsEqualIID(riid, &IID_IAudioClockAdjustment)) {
+        IAudioClockAdjustment_AddRef(&This->IAudioClockAdjustment_iface);
+        *ppv = &This->IAudioClockAdjustment_iface;
     } else if (IsEqualIID(riid, &IID_IAudioStreamVolume)) {
         IAudioStreamVolume_AddRef(&This->IAudioStreamVolume_iface);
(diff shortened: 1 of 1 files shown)


ivyl:
I've wrote test for this scenario and `GetService()` doesn't seem to work with this interface. The only other game using it I'm aware of, DOOM Eternal, uses `QueryInterface()`.

See: https://gitlab.winehq.org/wine/wine/-/merge_requests/5585/diffs?commit_id=6be262db9a156d176d8f82fca2fce2407afcedc3#0481926a1332eeb607de11080c6209bd5431f7c0_646_674

If you could write a test confirming that this is intended to work in some scenario on Windows that would be great!

Etaash-mathamsetty:
> I've wrote test for this scenario and `GetService()` doesn't seem to work with this interface. The only other game using it I'm aware of, DOOM Eternal, uses `QueryInterface()`.
> 
> See: [gitlab.winehq.org/wine/wine/-/merge_requests/5585/diffs?commit_id=6be262db9a156d176d8f82fca2fce2407afcedc3](https://gitlab.winehq.org/wine/wine/-/merge_requests/5585/diffs?commit_id=6be262db9a156d176d8f82fca2fce2407afcedc3#0481926a1332eeb607de11080c6209bd5431f7c0_646_674)
> 
> If you could write a test confirming that this is intended to work in some scenario on Windows that would be great!

you need AUDCLNT_STREAMFLAGS_RATEADJUST in IAudioClient_Initialize for it to work I guess, but there's no easy way to check for that with the current code in proton so just return the interface and give the app the benefit of a doubt that they did actually pass in the flag.

ivyl:
Thanks! I'll sort this out and add the appropriate checks to the upstreaming attempt.

ivyl:
https://github.com/ValveSoftware/wine/commit/486bd18f7c518bc105d1a5aa8ffdfe4bd332e7a7

PR #256, opened 2024-09-10
fsync: Fix checking if futex_waitv is available.

mati7337:
Proton's wine checks if futex_waitv syscall is available by checking if it sets errno to ENOSYS.
This works in most cases, however in environments like podman even though this syscall is blocked it sets errno to EPERM instead which causes fsync to be enabled despite it not being supported. This causes 100% CPU usage even when launching notepad.exe.

In podman this seems to be a problem since https://github.com/containers/common/commit/a3f335c70c37639e1ee8f2f613414ee24d0ce814 which added futex_waitv to the seccomp blocklist. It seems that before that commit it returned ENOSYS which wine correctly recognized as unsupported.

This PR fixes this by additionally checking if errno isn't EPERM before enabling fsync.


ivyl:
Indeed, looks like there's no other way we can accidentally get EPERM from this call other than seccomp. It's in bleeding-edge and reach all the other Proton versions from there. Thank you.

PR #257, opened 2024-09-25
Fixes for Electra Media Player (Unreal Engine plugin)

nsivov:


ivyl:
Closing since 8.0 shouldn't see any development and there's no plan for a new release.

PR #262, opened 2024-11-03
kernelbase: Fix return on failed VirtualQueryEx

nulleve:
Issues mentioned in [#8208](https://github.com/ValveSoftware/Proton/issues/8208);

Returning specific booleans such as `FALSE`, `TRUE` causes `WriteProcessMemory` to break, noticeably with Dark and Darker's anti-cheat "TavernWorker".

This patch returns the `NTSTATUS` of `NtWriteVirtualMemory` as solely handled previously, however now only in cases where `VirtualQueryEx` responds with an error.

### DIFF
--- a/dlls/kernelbase/memory.c
+++ b/dlls/kernelbase/memory.c
@@ -613,7 +613,7 @@ BOOL WINAPI DECLSPEC_HOTPATCH WriteProcessMemory( HANDLE process, void *addr, co
     if (!VirtualQueryEx( process, addr, &info, sizeof(info) ))
     {
         close_cross_process_connection( list );
-        return FALSE;
+        return set_ntstatus( NtWriteVirtualMemory( process, addr, buffer, size, bytes_written ));
     }
 
     switch (info.Protect)


dlls/kernelbase/memory.c
```diff
@@ -613,7 +613,7 @@ BOOL WINAPI DECLSPEC_HOTPATCH WriteProcessMemory( HANDLE process, void *addr, co
     if (!VirtualQueryEx( process, addr, &info, sizeof(info) ))
     {
         close_cross_process_connection( list );
-        return FALSE;
+        return set_ntstatus( NtWriteVirtualMemory( process, addr, buffer, size, bytes_written ));
```
nsivov, line 616: This does not look right. If you want to have the last error set, you'll need to use NtQueryVirtualMemory() and use status from that. Note that set_ntstatus() will still return either 0 or 1.

Also, since part is not specific to proton, please submit the fix together with some tests replicating failure you see, at https://gitlab.winehq.org/wine/wine.

nulleve: Ah I see, I saw the commits in this repo without checking if it was in wine's repo sorry about that. I attempted to use `NtQueryVirtualMemory` however I'm not familiar with windows internals (let alone their kernel), so while it compiles/runs it doesn't produce the same working effect as `NtWriteVirtualMemory` (TavernWorker in Dark and Darker still fails to run) and I don't have insight as to why to be honest.

Thank you for the insight and help, I'll try and send this over to wine's repo and see if someone is able to resolve this.

PR #265, opened 2024-12-09
win32u: HACK: Hide a window also for Post Mortem.

eliadevito:
Apply same hack of syberia to Post Mortem to fix black screen like [f131c0c](https://github.com/ValveSoftware/wine/commit/f131c0c65b2ff77d4032cfe2946748552ba8a665)

### DIFF
--- a/dlls/win32u/window.c
+++ b/dlls/win32u/window.c
@@ -479,7 +479,7 @@ HWND WINAPI NtUserSetParent( HWND hwnd, HWND parent )
         WCHAR name[32];
         UNICODE_STRING us = { 0, sizeof(name), name };
 
-        if (NtUserGetClassName( hwnd, FALSE, &us ) && !wcscmp( us.Buffer, u"SyberiaRenderWindowClass" ))
+        if (NtUserGetClassName( hwnd, FALSE, &us ) && ( !wcscmp( us.Buffer, u"SyberiaRenderWindowClass" ) || !wcscmp( us.Buffer, u"Post MortemRenderWindowClass" ) ))
         {
             ERR( "HACK: Hiding window.\n" );
             was_visible = FALSE;


kisak-valve:
Doubling up the preexisting if condition looks incorrect.

eliadevito:
> Doubling up the preexisting if condition looks incorrect.

Sorry... I made a mistake with the commit... fixed now

gofman:
Does the problem also happens on WIndows with the game? Otherwise, which is the actual problem which gets worked around this way (what doesn't work right without the hack)? Without understanding that (and exploring proper fix possibilities) we probably can't add such a hack just on the basis that it helps.

gofman:
I've got some additional info that it is the game from the same dev from the same time, looked at ddraw logs to at least confirm that it even uses ddraw in exclusive fullscreen mode (and also that the game seems totally broken on modern Windows). So probably the hack is fine in this case. Thanks!

eliadevito:
Yes, the problem also occurs on Windows.

Without this hack, the game is completely unplayable. The screen remains black, and during videos, it flashes constantly. Essentially, the game doesn't work correctly without it. I've tested and completed the entire game with the hack enabled, and everything functions properly when it's applied (with the exception of Alt+Tab, similar to the issue in Syberia).

eliadevito:
> I've got some additional info that it is the game from the same dev from the same time, looked at ddraw logs to at least confirm that it even uses ddraw in exclusive fullscreen mode (and also that the game seems totally broken on modern Windows). So probably the hack is fine in this case. Thanks!

Yes the game is from same dev and probably use the same engine

Regarding the Alt+Tab issue, I was able to workaround it using gamescope.

gofman:
> Without this hack, the game is completely unplayable.

Note that this alone is not an argument for introducing such sort of a hack. We normally find the actual underlying problem why it doesn't work like on Windows and fix the core reason, and only add hacks if that doesn't look immediately possible for some reason and after knowing which problem we are working around.

eliadevito:
Perfectly clear 😄. In my previous message, when I mentioned the game doesn't work without the hack, I wasn't advocating for it to be applied upstream for that reason. I was simply explaining the title's behaviour without it.

alasky17:
After some further investigation, we ended up pulling in this hack.  It seems clear that this is the same underlying issue because it is published by the same game dev from the same timeframe.   @eliadevito Thank you for the PR!  It should show up in experimental-bleeding-edge soon :) 

Edit: 05b9f4b0faf415941b25b2846da539eef47ab83b

PR #269, opened 2025-02-03
Hid multi TLC and Fanatec wheel-bases hidraw white-list

gotzl:
This PR pulls upstream changes that implement multi TLC support (https://gitlab.winehq.org/wine/wine/-/merge_requests/6206) and adds a list of Fanatec wheel-bases for which HIDRAW should be enabled.

## Background
Several sim-racing games crash upon detecting a connected Fanatec wheel-base. Affected titles include:
- *Automobilista 2*  
- *Assetto Corsa EVO*  
- *EA Sports WRC*  
- *Rennsport*  

The root cause is that these games include the **FanatecSDK**, which provides advanced wheel functionality, such as LED and display support. However, the **FanatecSDK is incompatible** with the Windows input device derived from **libinput/SDL**. When the SDK detects a Fanatec wheel-base, it expects to find [HID multi-TLC](https://learn.microsoft.com/en-us/windows-hardware/drivers/hid/hidclass-hardware-ids-for-top-level-collections#case-2-single-function-device-with-multiple-tlc) registry paths, which do not exist. As a result, the code encounters an exception and crashes the game.

## Proposed Solution
To resolve this issue, this PR introduces **multi-TLC support** and enables **HIDRAW** for Fanatec wheel-bases. This approach ensures that:

- Windows input devices are created directly from the **HIDRAW descriptor**, so that all expected device paths exist.  
- The **FanatecSDK can communicate** with the HID device via HIDRAW, allowing it to control LEDs and displays properly.  

## Side Note
Additionally, I have modified the **experimental Fanatec wheel-base driver** to extend the **HIDRAW HID descriptor** with the required HID-PID components. This change enables **force feedback (FFB)** for the Windows input device.  

For further details, see:  
- https://github.com/gotzl/hid-fanatecff/pull/86
- [hid-fanatecff README](https://github.com/gotzl/hid-fanatecff?tab=readme-ov-file#integration-with-wineproton)

### DIFF
--- a/dlls/dinput/tests/hid.c
+++ b/dlls/dinput/tests/hid.c
@@ -730,7 +730,6 @@ void hid_device_stop( struct hid_device_desc *desc, UINT count )
     for (i = 0; i < count; ++i)
     {
         ret = WaitForSingleObject( device_removed, i > 0 ? 500 : 5000 );
-        todo_wine_if(i > 0)
         ok( !ret, "WaitForSingleObject returned %#lx\n", ret );
     }
 }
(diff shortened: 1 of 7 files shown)


ivyl:
Multi-TLC support was picked up in Proton 10 rebase. I have merged the other patch adapting it to the current style of overrides keeping your authorship.

It's available right now in [bleeding-edge](https://github.com/ValveSoftware/Proton/wiki/Proton-Versions#proton-bleeding-edge) and trickle from there to all the release version.

Can you give it a try to make sure it works? I don't have any Fanatec device. Please report back if there are any problems.

gotzl:
Thank you very much! I did a quick test and it generally seems to work. I have FFB and the advanced wheel functions work as well. But there are some issues
1) When using an old prefix, games may crash after switching to proton-experimental somewhere in the FanatecSDK. However, after clearing the list of enumerated HID devices (`ROTON_VERSION="Proton Experimental" protontricks -c "wine reg delete 'HKLM\System\CurrentControlSet\Enum\HID' /f" APPID`) all games I've tested so far work.
2) Input devices are mixed-up (I've separated wheel and pedal devices). In-game, axis movement is showing up under the wrong device, while in `joy.cpl` everything looks fine. Also, at least in AssettoCorsaEvo I can see devices that shouldn't be there...

I'll investigate a bit when I get the time, but the second point could be a general issue in wine/proton input device handling.

PR #274, opened 2025-03-20
ntdll: HACK: Enable WINE_ALERT_SIMULATE_SCHED_QUANTUM option for other IdeaFactory games using Orochi 4

Naitrate:
This should help with massive frametime spikes problems in the following games, since they share the same game engine as Mary Skelter 2:

- Mary Skelter Finale (2357400)
- Death end re;Quest (Application.exe) (990050)
- Death end re;Quest 2 (1266220)
- Megadimension Neptunia VIIR (774511)
- Neptunia Virtual Stars (1399840)
- Dragon Star Varnir (1062040)


ivyl:
Merged into bleeding-edge. It should proliferate to released Proton version from there.

We'll be testing this to see if there's really any difference and we may have follow up changes.

PR #283, opened 2025-05-27
winebus: Fix detection of devices with no axis.

JacKeTUs:
Cherry-picked from https://gitlab.winehq.org/wine/wine/-/merge_requests/8088 (already merged into upstream)

This patch fixes detection of devices without an axis.
This will tremendously help users with various devices like button-boxes or gear-shifters. Right now users are relying on creating virtual devices with dummy buttons, and that's not ideal :)

Closes https://github.com/ValveSoftware/Proton/issues/7427


rbernon:
Thanks, I've cherry-picked the change, it should be in future Proton 10 releases.

PR #284, opened 2025-06-03
wintrust: Initialize all cert fields in WINTRUST_AddCert.

twhitehead:
This pull request is a back port of [this upstream patch](https://gitlab.winehq.org/wine/wine/-/commit/c04dc136dd7a5aa679dcdd22535fd96e25426174) that just landed to fix [regression 58293](https://bugs.winehq.org/show_bug.cgi?id=58293) that broke binary certificate verification in some cases.

It is required to make Revive (the OculusVR to OpenVR compatability layer) work again. Without it, `LibOVR.dll` refuses to load `LibOVRRT32dll` as it does not verify. Thanks!


rbernon:
Merged, should be in next Experimental releases. Thanks!

PR #285, opened 2025-06-10
Update main.c for virpil devices

ChaosRifle:
changed virpil (VPC) to be a catchall to use hidraw, as this is the only thing all vpc devices want. 

vpc devices have varried button count, axis count, and even pid depending on user config. Some users may have multiple of a same, or even similar, device with identical pid, where some poorly programmed games require the user to have varried pid, thus prompting them to input their own, making it effectively random. This includes some generations of L/R sticks (common), or using multiple of the same button panel (less common).

Currently, all vpc devices are generic joystick devices that work with hidraw, and prefer it. if vpc theoretically ever released a xbox like controller, this should be an if-else to exclude the xinput device's specific pid, as no reason to change the pid would exist for that.

### DIFF
--- a/dlls/winebus.sys/main.c
+++ b/dlls/winebus.sys/main.c
@@ -529,13 +529,10 @@ static BOOL is_hidraw_enabled(WORD vid, WORD pid, const USAGE_AND_PAGE *usages, 
         if (pid == 0x0127) prefer_hidraw = TRUE; /* VKB-Sim Space Gunfighter L */
         break;
     case 0x3344:
-        /* comes with 31 buttons in the default configuration, or 128 max */
-        if ((buttons == 31) || (buttons == 128)) prefer_hidraw = TRUE;
-        /* users may have configured button limits, usually 32/50/64 */
-        if ((buttons == 32) || (buttons == 50) || (buttons == 64)) prefer_hidraw = TRUE;
-        /* if customized, arbitrary amount of buttons may be shown, decide by PID */
-        if (pid == 0x412f) prefer_hidraw = TRUE; /* Virpil Constellation ALPHA-R */
-        if (pid == 0x812c) prefer_hidraw = TRUE; /* Virpil Constellation ALPHA-L */
+        prefer_hidraw = TRUE; /* All Virpil devices */
+        /* Virpil devices have configurable VID, PID, button count, and axis count. There is no good reason to change VID, so the only valid point to grab is VID
+        currently (2025/jun/10), virpil only sells generic input devices that prefer to be HIDRAW. if they ever sold an xinput device, explicit support should be set for that alone using an if else statement to exclude it from the catchall.
+        PID is different for different generations, but must be manually changed by end users for certain games as no difference exists for L/R on some generations of devices, making arbitrary PID a commonplace. */
         break;
     case 0x03eb:
         /* users may have configured button limits, usually 32/50/64 */


dlls/winebus.sys/main.c
```diff
@@ -529,13 +529,10 @@ static BOOL is_hidraw_enabled(WORD vid, WORD pid, const USAGE_AND_PAGE *usages,
...
         break;
     case 0x3344:
-        /* comes with 31 buttons in the default configuration, or 128 max */
-        if ((buttons == 31) || (buttons == 128)) prefer_hidraw = TRUE;
-        /* users may have configured button limits, usually 32/50/64 */
-        if ((buttons == 32) || (buttons == 50) || (buttons == 64)) prefer_hidraw = TRUE;
-        /* if customized, arbitrary amount of buttons may be shown, decide by PID */
-        if (pid == 0x412f) prefer_hidraw = TRUE; /* Virpil Constellation ALPHA-R */
-        if (pid == 0x812c) prefer_hidraw = TRUE; /* Virpil Constellation ALPHA-L */
+        prefer_hidraw = TRUE; /* All Virpil devices */
```
kakra, line 532: The xinput argument is pretty easy to get along with:
```c
if (buttons > 16)
    prefer_hidraw = TRUE;
```

because by definition, xinput devices cannot have more than 16 buttons:
https://learn.microsoft.com/en-us/windows/win32/api/xinput/ns-xinput-xinput_gamepad

dlls/winebus.sys/main.c
```diff
@@ -529,13 +529,10 @@ static BOOL is_hidraw_enabled(WORD vid, WORD pid, const USAGE_AND_PAGE *usages,
...
+        /* Virpil devices have configurable VID, PID, button count, and axis count. There is no good reason to change VID, so the only valid point to grab is VID
```
kakra, line 533: Very long lines... ;-)

kakra (commented):
The xinput argument is pretty easy to get along with...

dlls/winebus.sys/main.c
ChaosRifle: this actually breaks pedals. its not uncommon for virpil devices to have below 16 buttons, as pedals with axis-to-button is normal. in the event vpc ever released an xinput device (and they have never shown any interest in doing so, so its doubtful), a simple pid would be best, as vpc devices can have 0-8 axis and 0-128 buttons, or even have 32-button splits in firmware using multiple virtual devices to be compatible with older games.

this would also be broken by specific grips for sticks too, like the tomcat grip.

kakra: Okay, if there are actually devices with 16 or fewer buttons, then we shouldn't check for this - or maybe look for the axes, too? I'm pretty sure my original patch was able to also supply the axes in case you'd need additional heuristic checks - not sure if that made it into Proton 10.

OTOH, there's a bug in the `PROTON_ENABLE_HIDRAW` check which should be fixed soon in Proton Experimental and may work around a lot of issues people currently have: A lot of complaints with Virpil came up with Proton 10.

ChaosRifle: yes, multiple pedals, the warbrd stick grip, and tomcat grip, all have less than 16 buttons, typically. 

we dont really need any safety though as all current virpil products work on hidraw afaik - if one works, they all should, as they are all handled the same as each-other. I also dont think going by axis is safe, because that is variable and any theoretical xbox controller (which doesnt, and probably never would, exist) would have some axis too. I only brought up the xinput comment as just-in-case it ever happened, a paper trail would exist for a maintainer if I were not around to patch it, in the event virpil released an xinput device.

generally, the issue with virpil devices is them being treated as xinput, and they never should, for any of them.

kakra: > generally, the issue with virpil devices is them being treated as xinput, and they never should, for any of them.

Yep... Happens to a lot of joysticks. And I'm not sure if that is an artifact of SDL trying to turn them into gamepads. Apparently, udev doesn't seem to have a different identifier for joysticks vs gamepads either, however it can identify wheels and/or pedals (because those could be buttonlees devices). Everything would be so much easier if udev could properly do all of that.

ChaosRifle: indeed. I am not super well versed in it, but my understanding is that unknown devices are more commonly cheap xinput devices, so wine defaults those as xinput when unknown, as its the most common use case, at the detriment of joystick hardware.

anecdotally, not all pedals will be buttonless - the F-14 Tomcat uses toe brakes as a push to talk for intercom of the pilot and rio, and the mi-24 hind also has some toe brake functions, so players that want a realistic experience will typically bind an axis-to-button in the firmware (so it shows up as a real button) on their pedals to get that binding to work correctly in dcs. having both toe brakes use axis to button is also useful for other games that dont support axis for certain bindings.

kakra: > my understanding is that unknown devices are more commonly cheap xinput devices

True. I'm the author of xpadneo, and there's at least one manufacturer which looks identical to Microsoft in the HID descriptor but the rumble programming is actually not compatible - and usually it's easy by looking at the MAC OUI but of course this particular manufacturer doesn't even use registered MAC OUIs, they are completely random and made up.

So sometimes you have to go with the reverse tactics and actually look at the genuine OUI - or PID in this case - different case, similar story...

ChaosRifle:
merged upstream

PR #292, opened 2025-07-28
Fix DualSense buttons while using bluetooth

AmionSky:
Since wine 10 while using DualSense controller via bluetooth the R1 button also activated the touchpad click. The issue was that the third byte of the button report was not getting moved when the report gets remapped.
More info: https://bugs.winehq.org/show_bug.cgi?id=56883

### DIFF
--- a/dlls/winebus.sys/main.c
+++ b/dlls/winebus.sys/main.c
@@ -670,11 +670,11 @@ static void process_hid_report(DEVICE_OBJECT *device, BYTE *report_buf, DWORD re
          * Extended #41 report:
          *   Prefix X  Y  Z  Rz  TriggerLeft  TriggerRight  Counter  Buttons[3] ...
          */
-        if (report->buffer[0] == 0x31 && report->length >= 11)
+        if (report->buffer[0] == 0x31 && report->length >= 12)
         {
             BYTE trigger[2];
 
-            memmove(report->buffer, report->buffer + 1, 10);
+            memmove(report->buffer, report->buffer + 1, 11);
             report->buffer[0] = 1; /* fake report #1 */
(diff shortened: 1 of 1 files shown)


ivyl:
Hi. Thanks for troubleshooting this. Have you tried to upstream the change? This sounds like something that should go into upstream Wine too.

AmionSky:
I have not. I'm not even sure how to.
The fix is from the linked wine bug (but I also updated the length check). But that bug report has been stale for almost a month. So I opened this in hopes that this simple fix gets merged that has been annoying me for a while.

ivyl:
I had a closer look and indeed seems like we are memmoving to try to get rid of the first byte in front of the report to apply the quirk. Since we are memmoving a byte too few the last one moved gets "duplicated". It contains a state of R1 and the extra duplicated version is in the spot where touchpad state lives.

It all seems to make sense and I'll verify it with a report descriptor from a DualSense, see if we don't have a similar issue with a DualShock 4 and open upstream MRs for that.

Thank you for bringing it to my attention :-)

ivyl:
https://gitlab.winehq.org/wine/wine/-/merge_requests/8680

I've tweaked the commit message a bit but kept your authorship. The sing-off is to denote that I made some modifications. I hope that's okay.

Thanks!

ivyl:
https://github.com/ValveSoftware/wine/commit/0177314e9e450c2f0a4d71d1ab7c55fd6f68b5fb

The fix is available in [bleeding-edge](https://github.com/ValveSoftware/Proton/wiki/Proton-Versions#proton-bleeding-edge) and will make it's way from there to Proton Experimental and the next stable release.

PR #311, opened 2026-01-17
Prevent XGenericEvents from leaking memory while Steam overlay is open

yepokay:
XCheckIfEvent calls _XStoreEventCookie which adds an XGenericEventCookie to a doubly-linked list. If we don't remove it with XGetEventData we end up with a memory leak.

More info in this comment: https://github.com/ValveSoftware/steam-for-linux/issues/11446#issuecomment-3764174787

### DIFF
--- a/dlls/winex11.drv/event.c
+++ b/dlls/winex11.drv/event.c
@@ -514,7 +514,12 @@ static BOOL process_events( Display *display, Bool (*filter)(Display*, XEvent*,X
     while (XCheckIfEvent( display, &event, filter, (char *)arg ))
     {
         count++;
-        if (overlay_enabled && filter_event( display, &event, (char *)overlay_filter )) continue;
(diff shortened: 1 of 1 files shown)


kisak-valve:
I'm not convinced this is related to the linked issue report, but ignoring that, @rbernon should be pinged for a fix up for https://github.com/ValveSoftware/wine/commit/3b7769fc4fe1830c84d6d6ab9e9e0311e09d7fe8.

yepokay:
You might be correct, it might just be one of multiple causes for these timebomb issues people have been having. The linked report just seemed the most relevant.

And to be clear, this bug was introduced in Proton 5.13: https://github.com/ValveSoftware/wine/commit/f649de43236f67943a44ae89dfbbd3d4c7d9eeb4

rbernon:
Thank you, I've merged it into the Proton Experimental, it should be available soon in the bleeding-edge branch.

PR #314, opened 2026-02-23
Add Logitech G29 wheel to custom mapping override

LuanVSO:
with this change all of the following games correctly map the g29 pedals with no manual intervention required:
1. dirt rally 2.0 (690790)
1. WRC 7 (621830)
1. FIA European Truck Racing Championship (940580)

also note that, before this change, games 2 and 3 required the file at `<gamefolder>/Common/Settings/InputDeviceConfig.lua` to be edited otherwise the game would not recognize the throttle and brake pedals

### DIFF
--- a/dlls/dinput/joystick_hid.c
+++ b/dlls/dinput/joystick_hid.c
@@ -49,6 +49,7 @@
 
 #define VID_LOGITECH 0x046D
 #define PID_LOGITECH_G920 0xC262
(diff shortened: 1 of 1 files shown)


LuanVSO:
Hi, could i please get a review on this?
i don't know if this would fit on upstream wine as it is based on the already existing proton patch for the g920

PR #316, opened 2026-03-01
fix(qcap,qedit): stabilize preview negotiation and allocator behavior for Sample Grabber

LilFishyChan:
Prefer configured sink media type during Sample Grabber media type enumeration.

Improve SmartTee preview path for RGB24/RGB32 compatibility and conversion.

Handle allocator not-committed condition with commit-and-retry to avoid stream teardown.

Provide preview buffer-size decision callback to prevent null callback crashes.


kisak-valve:
Hello @LilFishyChan, you've come across a downstream fork of wine used in Proton. This pull request should be sent to the upstream wine project for review. After it's been reviewed and merged, then we could ponder backporting it to Proton.

PR #317, opened 2026-03-22
Fix racy syscall entry breaking the Unity garbage collector

hoshinolina:
Upstream bug: https://bugs.winehq.org/show_bug.cgi?id=59333
Upstream MR: https://gitlab.winehq.org/wine/wine/-/merge_requests/10419

This is a backport to proton bleeding-edge, since that (or proton 10.0) is what the majority of users affected by this are running. I verified that the new ntdll:threadctx passes in Proton and that the ntdll:exception tests do not regress (those exercise the code affected by this).

I'm opening this here early for tracking and because I already did the backport (it's quite involved), and since most of the users affected by this are running some version of Proton 10/experimental.

### DIFF
--- a/dlls/ntdll/tests/Makefile.in
+++ b/dlls/ntdll/tests/Makefile.in
@@ -23,6 +23,7 @@ SOURCES = \
 	string.c \
 	sync.c \
 	thread.c \
+	threadctx.c \
 	threadpool.c \
 	time.c \
 	unwind.c \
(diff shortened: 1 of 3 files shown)


kisak-valve:
Hello @hoshinolina, this should be reviewed and merged upstream first. Once it's merged upstream, give me a ping here and I'll re-open this merge request.

PR #318, opened 2026-03-23
fix: ignore non-positional arguments for winebrowser except "app"

Ericky14:
I have found that sometimes developers use ShellExecute for example to manually launch browsers via winebrowser and might use browser arguments. Winebrowser doesn't support those and ends up forward the first one to the browser, such as `--new-window`.
With this change, winebrowser should be more resilient and find the correct URL to launch.

### DIFF
--- a/programs/winebrowser/main.c
+++ b/programs/winebrowser/main.c
@@ -421,6 +421,7 @@ int wmain(int argc, WCHAR *argv[])
     DWORD scheme;
     IUri *uri;
     HRESULT hres;
+    int i;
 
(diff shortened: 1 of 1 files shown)


kisak-valve:
Hello @Ericky14, you've come across a downstream fork of wine used in Proton. This merge request should be evaluated and merged into upstream wine before considering it for backport into Proton.

After it's upstream, please give me a ping and note the upstream commit and known affected game(s) so that there's a test case and a Proton dev can ponder this further.

PR #327, opened 2026-05-15
Add additional SGI check in window.c for Forza Horizon 6

galacticaledge:
This game seems to have the same issue as FH4 and FH5 where a black window shows by default. So adding a check can be done for FH6 too


ivyl:
Looks like this was already addressed by a4366f68620cdfb34926f8bba2619dab2ed1e34c.

Thanks!

PR #328, opened 2026-05-17
[Crash] win32u/vulkan: Move 'features' to function scope in vkCreateDevice.

nanomatters:
VkPhysicalDeviceFeatures features is declared inside an inner {} block
in win32u_vkCreateDevice, but its address is stashed into
create_info->pEnabledFeatures and then dereferenced after the block ends
by convert_device_create_info() and the host vkCreateDevice call. This
was exposed when compiled using gcc16 and caused game crashes.

Originated in c67d194e7c8 ("fshack: win32u: Support vulkan scaling
according to fake resolution.").

### DIFF
--- a/dlls/win32u/vulkan.c
+++ b/dlls/win32u/vulkan.c
@@ -895,6 +895,7 @@ static VkResult win32u_vkCreateDevice( VkPhysicalDevice client_physical_device, 
     unsigned int queue_count, props_count, i;
     struct vulkan_device *device;
     struct mempool pool = {0};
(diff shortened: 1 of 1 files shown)


ivyl:
This is now in [`bleeding-edge`](https://github.com/ValveSoftware/Proton/wiki/Proton-Versions#proton-bleeding-edge).

I've turned the commit into a `fixup!` one against the original so it's easier to pair them during the next rebase. I've also removed the unnecessary whitespace changes.

Thanks for the fix

PR #330, opened 2026-05-22
ntdll: Handle ARM64EC read and map notifications for copied code.

dllu:
Fixes Starcraft II x64 client on arm64 (NVIDIA DGX Spark), along with https://github.com/FEX-Emu/FEX/pull/5508 and https://github.com/FEX-Emu/FEX/pull/5510

BEFORE:

image

AFTER:

image
# Wine ARM64EC Read And Map Follow-up

This note describes a Wine-side ARM64EC fix needed by StarCraft II's x64 client
when running through ARM64 Proton on FEX.

## Summary

StarCraft II's x64 client exercises two ARM64EC cases where Wine has better
source information than the emulator:

- An executable non-image section view can replace an earlier image mapping, but
  the ARM64EC notification path only reports the image map.
- The process can read ARM64EC DLL bytes into a private buffer and execute copied
  syscall wrapper code from that buffer, bypassing the normal image loader path
  that applies ARM64X dynamic-value fixups.

The proposed fix handles both cases in Wine:

- report successful executable non-image section mappings through the existing
  ARM64EC memory-protection notification callback;
- apply ARM64X dynamic-value fixups to synchronous `NtReadFile` buffers when the
  source file is an ARM64EC image that contains the relevant metadata.

This prevents the emulator from having to infer missing Wine state later from an
execution-time fault or decode miss.

## Executable Non-image Section Maps

Wine already notifies ARM64EC support code when mapping image views. StarCraft II
also creates executable non-image section views that can occupy an address range
previously associated with an image view.

If that later mapping is not reported, the emulator may keep stale executable
range metadata for the old view. The fix is to extend the ARM64EC
`NtMapViewOfSection` notification path so successful non-image maps report their
final protection through the existing `pNotifyMemoryProtect()` callback.

The important properties are:

- notify only after `NtMapViewOfSection` succeeds;
- use the final mapped base, size, and protection returned by Wine;
- keep image-map notifications on the existing image path;
- use the memory-protection callback for non-image mappings rather than adding a
  new emulator-side reconstruction path.

## Copied ARM64EC Syscall Wrapper Bytes

Wine's normal ARM64EC image mapping path applies ARM64X dynamic-value
relocations while loading images. A plain file read does not go through that
loader path.

StarCraft II reads ARM64EC DLL bytes into a private buffer and later executes
copied syscall wrapper code from that buffer. Without applying the ARM64X
dynamic-value fixups to the copied bytes, the private copy can differ from the
code Wine would have produced for a mapped ARM64EC image.

The fix is to wrap ARM64EC `NtReadFile` so that, after a successful synchronous
read, Wine asks the Unix side to inspect the backing file and apply ARM64X
dynamic-value fixups to the read buffer when appropriate.

The implementation should:

- only run after successful synchronous reads with known byte ranges;
- inspect the source file as a PE image and locate the ARM64X dynamic-value
  relocation metadata;
- apply fixups only to records whose file offsets overlap the bytes returned to
  the caller;
- leave unrelated files and unrelated read ranges untouched;
- reuse Wine's existing ARM64X relocation concepts instead of hard-coding
  syscall thunk byte patterns.

## Files Touched In The Prototype

The prototype patch is limited to `ntdll`:

```text
dlls/ntdll/signal_arm64ec.c
dlls/ntdll/unix/file.c
dlls/ntdll/unix/loader.c
dlls/ntdll/unix/unix_private.h
dlls/ntdll/unix/virtual.c
dlls/ntdll/unixlib.h
```

The public ARM64EC syscall wrapper lives in `signal_arm64ec.c`. The Unix-side
file descriptor handling is in `file.c`. The PE/ARM64X inspection and relocation
application live in `virtual.c`, with small declarations and unixlib plumbing in
the remaining files.

## Why Not Fix This In FEX

Two emulator-side workarounds were considered:

- lazily rebuilding executable tracking after a missing executable range;
- rewriting copied Wine syscall thunks immediately before decode.

Both solve symptoms after Wine has already lost or bypassed the relevant state.
Wine knows the section mapping result, file descriptor, read offset, and ARM64X
metadata at the time the operations occur. Handling the cases in Wine keeps the
emulator interface explicit and avoids making FEX guess which mappings or copied
buffers came from ARM64EC Wine internals.

## Validation

With the Wine changes rebuilt into ARM64 Proton, and with the remaining FEX
privileged-instruction handling fix applied, StarCraft II's x64 client launches
through its 64-bit switcher and creates a normal full-size game window.

The Proton log still contains repeated handled illegal-instruction exceptions
from StarCraft II's protected startup probes after the window is alive. Those
exceptions are expected for this title and are not the startup failure that the
Wine changes address.

### DIFF
--- a/dlls/ntdll/signal_arm64ec.c
+++ b/dlls/ntdll/signal_arm64ec.c
@@ -33,6 +33,7 @@
 #include "wine/list.h"
 #include "ntdll_misc.h"
 #include "unwind.h"
+#include "unixlib.h"
 #include "wine/debug.h"
 #include "ntsyscalls.h"
 
@@ -723,9 +724,17 @@ static void notify_map_view_of_section( HANDLE handle, void *addr, SIZE_T size, 
     SECTION_IMAGE_INFORMATION info;
     NTSTATUS status;
 
+    if (NtQuerySection( handle, SectionImageInformation, &info, sizeof(info), NULL ))
+    {
+        if (pNotifyMemoryProtect)
+        {
+            pNotifyMemoryProtect( addr, size, protect, FALSE, 0 );
+            pNotifyMemoryProtect( addr, size, protect, TRUE, STATUS_SUCCESS );
+        }
+        return;
+    }
     if (!pNotifyMapViewOfSection) return;
     if (!NtCurrentTeb()->Tib.ArbitraryUserPointer) return;
-    if (NtQuerySection( handle, SectionImageInformation, &info, sizeof(info), NULL )) return;
     status = pNotifyMapViewOfSection( NULL, addr, NULL, size, alloc, protect );
     if (NT_SUCCESS(status)) return;
     NtUnmapViewOfSection( GetCurrentProcess(), addr );
@@ -806,17 +815,34 @@ NTSTATUS SYSCALL_API NtReadFile( HANDLE handle, HANDLE event, PIO_APC_ROUTINE ap
                                  IO_STATUS_BLOCK *io, void *buffer, ULONG length,
                                  LARGE_INTEGER *offset, ULONG *key )
 {
+    struct apply_arm64x_read_fixups_params params =
+    {
+        .handle = handle,
+        .buffer = buffer,
+        .offset = offset && offset->QuadPart >= 0 ? offset->QuadPart : -1,
+    };
     NTSTATUS status;
 
     if (pBTCpu64NotifyReadFile && enter_syscall_callback())
     {
         pBTCpu64NotifyReadFile( handle, buffer, length, FALSE, 0 );
         status = syscall_NtReadFile( handle, event, apc, apc_user, io, buffer, length, offset, key );
+        if (!status && io->Information)
+        {
+            params.size = io->Information;
+            WINE_UNIX_CALL( unix_apply_arm64x_read_fixups, &params );
+        }
         if (pBTCpu64NotifyReadFile) pBTCpu64NotifyReadFile( handle, buffer, length, TRUE, status );
         leave_syscall_callback();
         return status;
     }
-    return syscall_NtReadFile( handle, event, apc, apc_user, io, buffer, length, offset, key );
+    status = syscall_NtReadFile( handle, event, apc, apc_user, io, buffer, length, offset, key );
+    if (!status && io->Information)
+    {
+        params.size = io->Information;
+        WINE_UNIX_CALL( unix_apply_arm64x_read_fixups, &params );
+    }
+    return status;
 }
 
 NTSTATUS SYSCALL_API NtSetContextThread( HANDLE handle, const CONTEXT *context )
--- a/dlls/ntdll/unix/file.c
+++ b/dlls/ntdll/unix/file.c
@@ -6728,6 +6728,25 @@ err:
 
 
 /******************************************************************************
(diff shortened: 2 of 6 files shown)


dllu:
@bylaws after your comment in https://github.com/FEX-Emu/FEX/pull/5509, I made this PR. I think that you're right that the "fix" in FEX is a gross workaround.

> this is wrong - but i get the vibe you're going for. If you wanna make a c program that reprooduces i can fix it in a cleaner way, but I can probably figure out from this issue.

I will admit that I am new to this and relied on AI to generate the fix.

Here's a reproducing C program: https://gist.github.com/dllu/b0277fe3165578951d99c3c4af56bca0 

## Build

From a directory containing `scripts/fex_exec_section_remap_repro.c`:

```bash
mkdir -p repros

docker run --rm \
  -u "$(id -u):$(id -g)" \
  -v "$PWD:/work" \
  -w /work \
  registry.gitlab.steamos.cloud/proton/steamrt4/sdk/arm64-llvm:4.0.20260331.220802-0 \
  /opt/llvm-mingw/bin/x86_64-w64-mingw32-clang \
    -Wall -Wextra -O2 -g \
    -o repros/fex_exec_section_remap_repro.exe \
    scripts/fex_exec_section_remap_repro.c
```

## Run

Set these paths for your environment:

```bash
PROTON=/path/to/arm64-proton/dist/proton
PREFIX=/path/to/temporary/compatdata
REPRO="$PWD/repros/fex_exec_section_remap_repro.exe"
OUT=/tmp/fex_exec_section_remap_repro.out
LOGDIR=/tmp/fex-section-remap-repro-logs
APPID=123456
```

`APPID` only needs to be a stable placeholder for Proton's environment. The
repro does not depend on Steam, a game install, or a specific Wine prefix.

```bash
rm -f "$OUT"
mkdir -p "$LOGDIR"

env \
  STEAM_COMPAT_APP_ID="$APPID" \
  SteamAppId="$APPID" \
  SteamGameId="$APPID" \
  STEAM_COMPAT_CLIENT_INSTALL_PATH="$HOME/.steam/steam" \
  STEAM_COMPAT_DATA_PATH="$PREFIX" \
  PROTON_USE_ARM64=1 \
  PROTON_LOG=1 \
  PROTON_LOG_DIR="$LOGDIR" \
  WINEDEBUG=-all \
  "$PROTON" runinprefix "$REPRO" 'Z:\tmp\fex_exec_section_remap_repro.out'

cat "$OUT"
```

## Expected Results

With Wine reporting executable non-image section maps to the ARM64EC/FEX memory
notification path, the second call returns:

```text
section executable page: calling 0000000040F70000, expected 0x22222222
section executable page: returned 0x22222222
failures=0
```

Without that Wine fix, the program exits nonzero after the second call starts:

```text
section executable page: base=0000000040F70000 ... protect=0x40 type=0x40000
section executable page: calling 0000000040F70000, expected 0x22222222
```

The Proton/FEX log shows the missing executable range:

```text
NoExec instruction in entry block: 40F70000
Exception: Code: C0000005 Address: 40F70000
```

kisak-valve:
Hello @dllu, you've come across a downstream fork of wine used with Proton. This change should be evaluated and merged upstream first, and after that, make a note here pointing to the merged commit and a Proton dev can consider backporting it to Proton.

PR #331, opened 2026-05-24
belauncher: Prepend install path for game_exe starting with backslash.

PhialsBasement:
`belauncher.exe` fails to launch the game when a title's `BELauncher.ini` specifies a `64BitExe=` value with a leading backslash (a Win32 "root-relative" path), e.g.:

```ini
64BitExe=\Game\Binaries\Win64\Game.exe
```

The existing guard in `programs/belauncher/main.c` only prepends the install directory when `PathIsRelativeW(game_exeW)` returns TRUE, however, `PathIsRelativeW` returns FALSE for any path whose first character is `\` (per the implementation in `dlls/kernelbase/path.c`)

Reproduced under Proton Experimental (11.0-20260518b) with a title whose `BELauncher.ini` uses a leading-backslash `64BitExe=`. With `WINEDEBUG=+belauncher`:

```
trace:belauncher:wWinMain PROTON_ORIG_LAUNCHER_NAME L"S:\common\<Game>\<Game>.exe".
trace:belauncher:wWinMain cmdline L"-XRRuntime=... -d3d11".
trace:belauncher:wWinMain Launching original launcher L"S:\common\<Game>\<Game>.exe".
trace:belauncher:wWinMain Launching game executable \<Game>\Binaries\Win64\<Game>.exe for BattlEye.
trace:belauncher:wWinMain game_exe L"\<Game>\Binaries\Win64\<Game>.exe".
trace:belauncher:wWinMain path L"S:\common\<Game>\", be_arg (null).
trace:belauncher:wWinMain launch_cmd L"\<Game>\Binaries\Win64\<Game>.exe -XRRuntime=... -d3d11".
err:belauncher:wWinMain CreateProcessW failed, error 2.
```

Note `launch_cmd` is missing the `path` prefix, even though `path` is known and traced one line earlier. The file at `S:\common\<Game>\<Game>\Binaries\Win64\<Game>.exe` exists; the path Wine actually queried (`<current_drive>:\<Game>\Binaries\Win64\<Game>.exe`) does not.

Editing the title's `BELauncher.ini` to remove the leading backslash makes the launch succeed, confirming the diagnosis.

P.S this fixes Contractors Showdown/Exfilzone which is yet another W for the steam frame

### DIFF
--- a/programs/belauncher/main.c
+++ b/programs/belauncher/main.c
@@ -169,7 +169,16 @@ int WINAPI wWinMain(HINSTANCE hInst, HINSTANCE hPrevInst, LPWSTR cmdline, int cm
     battleye_status = 0x9; /* Launching Game */
     _write(1, &battleye_status, 1);
 
-    if (PathIsRelativeW(game_exeW) || game_exeW[0] == L'\\')
+    /* Windows BattlEye treats a singleton leading-backslash 64BitExe as
+     * install-relative; strip it so the joined launch_cmd doesn't end up
+     * with a doubled separator.  Skip if a second '\' follows so UNC paths
+     * are preserved. */
+    if (game_exeW[0] == L'\\' && game_exeW[1] != L'\\')
+    {
+        memmove(game_exeW, game_exeW + 1, game_exe_len * sizeof(*game_exeW));
+        --game_exe_len;
+    }
+    if (PathIsRelativeW(game_exeW))
         path_len = wcslen(path);
     else
         path_len = 0;
--- a/programs/belauncher/main.c
+++ b/programs/belauncher/main.c
@@ -169,7 +169,7 @@ int WINAPI wWinMain(HINSTANCE hInst, HINSTANCE hPrevInst, LPWSTR cmdline, int cm
     battleye_status = 0x9; /* Launching Game */
     _write(1, &battleye_status, 1);
 
-    if (PathIsRelativeW(game_exeW))
+    if (PathIsRelativeW(game_exeW) || game_exeW[0] == L'\\')
         path_len = wcslen(path);
     else
         path_len = 0;


programs/belauncher/main.c
```diff
@@ -169,7 +169,7 @@ int WINAPI wWinMain(HINSTANCE hInst, HINSTANCE hPrevInst, LPWSTR cmdline, int cm
     battleye_status = 0x9; /* Launching Game */
     _write(1, &battleye_status, 1);
 
-    if (PathIsRelativeW(game_exeW))
+    if (PathIsRelativeW(game_exeW) || game_exeW[0] == L'\\')
```
Hi-Angel: I think it might be useful to add a comment about `game_exeW[0] == L'\\'`.

I'm not a Windows dev, but I searched around and it seems leading backslash is supposed to translate to the root of the disk we're currently in.

That implies the line `game_exeW[0] == L'\\'` is working around BattleEye bug, where it translates the dir incorrectly *(i.e. the game that works on Windows actually wasn't supposed to work there either)*.

So, I think it's useful to add a comment explaining the situation, i.e. why exactly we do the `game_exeW[0] == L'\\'` check here.

BananaWorks07:
Maybe it would be better to check if the path is relative on the string + 1 to remove the \ character?

Something like 
```
If (*str = L'\\') str++;
```

programs/belauncher/main.c
PhialsBasement: Ive just applied some comments to help identify that.

PhialsBasement:
> Maybe it would be better to check if the path is relative on the string + 1 to remove the \ character?
> 
> Something like
> 
> ```
> If (*str = L'\\') str++;
> ```

Good idea, done with memmove since game_exeW is an array rather than a pointer, plus a `game_exeW[1] != L'\\'` guard so UNC paths aren't mangled.

BananaWorks07:
I think using memmove is a bit overkill, just define a second variable that points to the array

PhialsBasement:
> I think using memmove is a bit overkill, just define a second variable that points to the array

I'd prefer to keep the memmove, it leaves `game_exeW` in a single consistent state after the strip, so later code (here or any future patch) can't go for the original array and get the unstripped form.

PhialsBasement:
Hi @kisak-valve, would you mind putting this one in front of the developers for a quick look when there is a chance? It fixes Contractors Showdown / Exfilzone, which has had BattlEye enabled for Proton for about a year but has been unable to launch that whole time because of the leading backslash in its BELauncher.ini. Since belauncher only lives in this tree there is no upstream merge request to point at. Happy to make any changes if something else is wanted.

PR #332, opened 2026-06-01
Add aligment fix for 4:2 planar video

Fktor:
Issue: https://github.com/valvesoftware/proton/issues/576

### DIFF
--- a/dlls/winegstreamer/wg_parser.c
+++ b/dlls/winegstreamer/wg_parser.c
@@ -126,7 +126,7 @@ struct wg_parser_stream
     GstBuffer *buffer;
     GstMapInfo map_info;
 
-    bool flushing, eos, enabled, has_tags, has_buffer, no_more_pads, fix_nv12;
+    bool flushing, eos, enabled, has_tags, has_buffer, no_more_pads, fix_align;
 
     uint64_t duration;
     gchar *tags[WG_PARSER_TAG_COUNT];
@@ -739,7 +739,7 @@ static gboolean sink_event_cb(GstPad *pad, GstObject *parent, GstEvent *event)
         {
             GstStructure *structure;
             GstVideoInfo video_info;
-            bool fix_nv12 = false;
+            bool fix_nv12 = false, fix_yv12 = false;
             GstCaps *caps;
 
             gst_event_parse_caps(event, &caps);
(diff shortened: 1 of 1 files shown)


Fktor:
Trying to fix a Resident Evil 0 video issue. The problem is that we have a video in 4:2:0 format, and the uv planes have a width that is not a multiple of 4. Gstreamer forces the stride to be aligned to 4, while the game expects the original stride. This causes visual green/red line artifacts.

I've attached the fix to a NV12 fix that was already present since they're from the same family of problems.

First time trying to fix something in Proton, so give it a hard look.

cmccarthy-codeweavers:
Should be good after some minor fixes. This loop:

```
        dst = map_info.data + dst_info.offset[plane1];
        src = map_info.data + src_info.offset[plane1];
        for (i = 0; i < aligned_height; ++i)
        {
            memmove(dst, src, dst_info.stride[plane1]);
            dst += dst_info.stride[plane1];
            src += src_info.stride[plane1];
        }
```

The height is the same for plane 2, so instead of being duplicated for plane2, it can be inside an outer loop: `for (plane = 1; plane <=2; ++plane)`

`else` should occur on a new line instead of `} else`

Extra condition lines are always only double-indented (4 spaces x 2), so it would be:
```
                fix_yv12 = (GST_VIDEO_INFO_FORMAT(&video_info) == GST_VIDEO_FORMAT_YV12
                        || GST_VIDEO_INFO_FORMAT(&video_info) == GST_VIDEO_FORMAT_I420)
                        && (video_info.stride[1] > GST_ROUND_UP_2(video_info.width/2));
```

Trace messages should mention I420 too, so it doesn't look like a format snafu occurred, e.g. something like `"Enabling the YV12/I420 alignment fix."`

I think it should be ok to just call `assert(0)` instead of `GST_ERROR("fix_align called for unsupported format.")`

The commit message should be something like:

`winegstreamer: HACK: Add alignment fix for 4:2 planar video.`

Fktor:
Thanks for the input. I've included the suggestions as well as adding a sanity assert for the size of the rewritten buffer.

cmccarthy-codeweavers:
Thanks for revising; I have pushed this to our experimental 11 branch. It will be included in the next bleeding edge 11 release. This PR can be closed.

PR #333, opened 2026-06-06
Fix black bar in A Plague Tale Requiem

adolfintel:
The game creates a window with borders, then removes them, leaving a black bar visible on top of the screen that could be removed by pressing Alt+Enter twice.
This patch detects the game using the SteamGameId environment variable and makes wine fix the position automatically.

Disclaimer: the issue was discovered and fixed by an LLM, but it works, so I 'm submitting the patch.


ivyl:
https://gitlab.winehq.org/wine/wine/-/wikis/Developer-FAQ#can-i-contribute-code-generated-by-a-large-language-model-llm-tool

This change is un-upstremable and impossible to work with.

adolfintel:
Understandable. I'll keep the patch in my repo in case anyone needs it.

PR #335, opened 2026-06-23
MSFS multiline text repeating workaround

rakslice:
This is the fix proposed by Tom Brazier in the 2019-09-11 15:12:38 UTC comment on https://bugs.winehq.org/show_bug.cgi?id=43493


rakslice:
This is re: https://github.com/ValveSoftware/Proton/issues/9902

kisak-valve:
Hello @rakslice, this should be reviewed and merged into upstream wine first. After that, then leave a note here with the commit and a Proton dev can ponder backporting it.

rakslice:
which was consolidated with https://github.com/ValveSoftware/Proton/issues/352

rakslice:
ok, cheers.

PR #336, opened 2026-06-30
cfgmgr32: Fix CM_Get_Device_ID buffer size handling.

nanomatters:
CM_Get_Device_ID_ExW() reserved no room for the terminating null in its
size check, which might overwrite outside the buffer bounds.

CM_Get_Device_ID_ExA() reused the caller's ANSI buffer length for
its temporary wide buffer and converted that full length back. Check the
size of the wide buffer from the actual id length and the ANSI output
before copying it to the caller's buffer.


kisak-valve:
Hello @nanomatters, you've come across a downstream fork of wine used in Proton. This should be reviewed and merged into the upstream wine project.

nanomatters:
@kisak-valve this pr actually fixes two known game crashes I referenced above. does it still have to go through wine?

kisak-valve:
Generally, yes, anything that can go upstream should go upstream first. After that, leave a note here pointing to the commit and a Proton dev can ponder backporting it into Proton.

rbernon:
I've created https://gitlab.winehq.org/wine/wine/-/merge_requests/11287 on your behalf, with some tests and corresponding changes to the code, I'll cherry-pick the change back into Proton after it's merged. Thanks for the patch!

nanomatters:
Thank you Rémi.

PR #340, opened 2026-07-13
winebus: Enable hidraw by default for Winwing Orion controllers.

hexontos:
Winebus may expose Winwing Orion flight hardware through SDL's synthesized gamepad path, which can produce an XInput-compatible device and hide axes or buttons available through the original HID reports.

This happened to me with (only) Orion Joystick (rest was okay) for Star Wars Squadron game making it bugged controller. I had to go to Protontricks and change protons aggressive remaping by forcing WINEBUS\WINE_COMP_HID:

```
Service: Changed from xinput to winehid
DeviceDesc: Changed to Wine HID compatible device
CompatibleIds: Changed to WINEBUS\WINE_COMP_HID (was already there, but first line was something else)
```
That fixed the issue, so it's also verified by me. This change uses winebus's existing hidraw preference mechanism so the raw HID interface wins during device arbitration.

The VID:PID match also applies to every interface exposed by these composite USB devices.

```
4098:bea8 — Winwing Orion Joystick Base Metal 2
4098:bd64 — Winwing Orion Throttle Base II
4098:bef0 — Winwing Orion Combat Rudder Pedals
```

### DIFF
--- a/dlls/winebus.sys/main.c
+++ b/dlls/winebus.sys/main.c
@@ -654,6 +654,11 @@ static BOOL is_hidraw_enabled(WORD vid, WORD pid, const USAGE_AND_PAGE *usages, 
         if ((buttons == 32) || (buttons == 50) || (buttons == 64)) prefer_hidraw = TRUE;
         if (pid == 0x2055) prefer_hidraw = TRUE; /* ATMEL/VIRPIL/200325 VPC Throttle MT-50 CM2 */
         break;
+    case 0x4098:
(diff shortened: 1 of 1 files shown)


kisak-valve:
Hello @hexontos, you've come across a downstream fork of wine used in Proton. This merge request should be reviewed and merged upstream first. After that, add a note here with a link pointing to the commit and a Proton dev can ponder backporting it here.

hexontos:
Okay thx, sorry for spam. I hopefully found correct upstream and reported it here https://bugs.winehq.org/show_bug.cgi?id=60012

PR #344, opened 2026-07-31
Image base high entropy

ghost:
Wine only asks the server for a shared image map address, and only maps top-down,
for DLLs. Executables are mapped at their preferred `ImageBase`, or wherever the
bottom-up search finds room.

Windows relocates ASLR-enabled executables like any other image, and places
`HIGH_ENTROPY_VA` ones high in the address space (`0x7ff6…`–`0x7ff8…`). Rainbow Six
Siege (appid 359550) sets both `DYNAMIC_BASE` and `HIGH_ENTROPY_VA`, and its
anti-tamper layer depends on landing there. Mapped at its preferred base
`0x140000000` it dereferences a non-canonical pointer and crashes seconds after
startup:

```
EXCEPTION_ACCESS_VIOLATION, read at 0xFFFFFFFFFFFFFFFF
RIP = RainbowSix.exe+0x5F35B64, RCX = 0xD2729D1609B37359
```

This lets executables use the same shared-map-address path as DLLs, and moves the
64-bit image pool into the range Windows uses. The image then maps at `0x7ff7…` and
the game runs. Reproduced on Proton 11.0, Experimental, Hotfix, GE-Proton10 and
proton-cachyos — all crash identically, all fixed by this change.

One caveat: the pool now spans ~8 GB instead of 16 TB, since `0x7ff600000000` sits
just below the user-space limit. Keying the placement on `HIGH_ENTROPY_VA` instead of
dropping the `IMAGE_FILE_DLL` test would narrow the behaviour change — happy to
rework it that way.

### DIFF
--- a/dlls/ntdll/unix/virtual.c
+++ b/dlls/ntdll/unix/virtual.c
@@ -3570,8 +3570,7 @@ static NTSTATUS map_image_view( struct file_view **view_ret, struct pe_image_inf
     void *base;
     NTSTATUS status;
     ULONG_PTR start, end;
-    BOOL top_down = (image_info->image_charact & IMAGE_FILE_DLL) &&
-                    (image_info->image_flags & IMAGE_FLAGS_ImageDynamicallyRelocated);
+    BOOL top_down = (image_info->image_flags & IMAGE_FLAGS_ImageDynamicallyRelocated);
 
     limit_low = max( limit_low, (ULONG_PTR)address_space_start );  /* make sure the DOS area remains free */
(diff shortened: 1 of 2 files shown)


kisak-valve:
Hello @xeralin2, you've come across a downstream fork of wine used in Proton . This pull request should evaluated and merged upstream first, then add a comment here with the commit and a Proton dev can ponder a backport.

ghost:
Thanks @kisak-valve 

Here is the merge request: https://gitlab.winehq.org/wine/wine/-/merge_requests/11547

PR #345, opened 2026-08-03
windows.ui.core.textinput: Return success from CoreInputView Try* methods

IceSqueez:
### Problem

Unity 6 titles (e.g. Albion Online, appid 761890) crash the moment a text field is committed (pressing Enter in chat, market search, etc.). The engine calls `CoreInputView.GetForCurrentView().TryHide()`; the stubs in `windows.ui.core.textinput` return `E_NOTIMPL`, which the C++/WinRT projection turns into an unhandled `hresult_not_implemented` exception:

```
fixme:coreinputview:core_input_view3_TryHide iface ..., result ... stub!
fixme:combase:RoOriginateLanguageException 0x80004001
trace:seh:dispatch_exception code=e06d7363 (EXCEPTION_WINE_CXX_EXCEPTION)
wine: Unhandled exception 0xe06d7363 ...
```

Disabling the DLL does not help either the activation failure takes the same unhandled-exception path. On Windows this API always exists and succeeds, so games have no reason to handle failure.

### Fix

1. Return success from the `TryShow*`/`TryHide*` methods: `TryShow*` reports `FALSE` (no pane was shown), `TryHide*` reports `TRUE` (nothing is shown, so it is hidden) same convention as `inputpane2_TryShow` in `windows.ui`.
2. Forward the calls to the tabtip window (`WM_TABTIP_OSK_TOGGLE`), mirroring the existing `InputPane` implementation, so the Steam on-screen keyboard follows text-field focus in CoreInputView-based titles. `TryShowWithKind` only toggles the keyboard for the `Default` and `Keyboard` kinds.

### Testing

Tested with Albion Online on Arch/Hyprland: without the patch the client crashes on every text-field commit; with it chat and market search work.
The OSK toggling follows the same mechanism as the existing InputPane implementation.

### DIFF
--- a/dlls/windows.ui.core.textinput/Makefile.in
+++ b/dlls/windows.ui.core.textinput/Makefile.in
@@ -1,5 +1,5 @@
 MODULE  = windows.ui.core.textinput.dll
-IMPORTS = combase
+IMPORTS = combase user32
 
 SOURCES = \
 	classes.idl \
--- a/dlls/windows.ui.core.textinput/main.c
+++ b/dlls/windows.ui.core.textinput/main.c
@@ -23,6 +23,8 @@
 
 WINE_DEFAULT_DEBUG_CHANNEL(coreinputview);
 
+#define WM_TABTIP_OSK_TOGGLE (WM_USER + 1)
+
 struct core_input_view
 {
     ICoreInputView ICoreInputView_iface;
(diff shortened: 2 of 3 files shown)


kisak-valve:
Hello @IceSqueez, you've come across a downstream fork of wine used in Proton. This merge request should be evaluated and merged upstream first. After that's happened, then add a comment here with the commit and a Proton dev can ponder a backport.

PR #346, opened 2026-08-06
winmm: Open mmio files using the Unicode file APIs

mrq1911:
Sound effects are completely silent in Steam app 2006210 (血染め鬼女) under Proton: every `mciSendString("open \"data/datN.w\" alias OTON.M type waveaudio")` fails with `MCIERR_FILE_NOT_FOUND` (172 of them at startup), while the MP3 music — `type MPEGVideo` → `mciqtz32` → `quartz` — plays normally.

### Cause

- `mciwave` opens files with `mmioOpenW()`, whose only file primitive is winmm's `create_file_OF()`.
- `create_file_OF()` resolved relative names with `SearchPathA()`, which is `SearchPathW()` followed by a lossy `RtlUnicodeStringToAnsiString()` of the **result**.
- The game's install directory is named `血染め鬼女`, which is not representable in the prefix's ANSI codepage (1252 here), so the correctly-resolved absolute path came back as `?????` and `CreateFileA()` failed with `ERROR_INVALID_NAME`.
- `mmio` maps any open failure to `MMIOERR_FILENOTFOUND`, which `mciwave` then reports as `can't find file` — misleading, since the file is present and readable.

Visible directly in a `+mmio,+file` trace, where `SearchPathW` succeeds and `CreateFileA` is handed the mangled result:

```
trace:file:SearchPathW found L"Z:\\...\\common\\\8840\67d3\3081\9b3c\5973\\data\\dat1.w"
trace:file:CreateFileW L"Z:\\...\\common\\?????\\data\\dat1.w" GENERIC_READ FILE_SHARE_READ creation 3
warn:file:NtCreateFile L"\\??\\Z:\\...\\common\\?????\\data\\dat1.w" not found (c0000033)
```

The filename the application passes is plain ASCII (`data/dat9.w`); only the *resolved* path cannot survive the round trip. `mciqtz32`/`quartz` are unaffected because they use `CreateFileW` and never touch `mmio`.

### Scope

- Non-ASCII **filenames** passed to `mmioOpenW()` are still mangled, by the `CP_ACP` conversion at the top of `mmioOpenW()` itself. Fixing that means plumbing WCHAR through `MMIO_Open()`/`mmioDosIOProc()`, which changes what custom ANSI IOProcs registered via `mmioInstallIOProcA()` receive. Deliberately not attempted here.
- The heap buffer rather than a fixed one is to avoid introducing a `MAX_PATH` limit on the `OF_CREATE` branch, which previously passed `path` straight to `CreateFileA()` unbounded.

### Testing

- New `winmm:mmio` test creates a directory whose name is outside the current ANSI codepage, chdirs into it, and opens a plain-ASCII relative filename with the same flags `mciwave` uses (`MMIO_ALLOCBUF | MMIO_DENYWRITE | MMIO_READ`). It fails with `0x101` (`MMIOERR_FILENOTFOUND`) before this change and passes after.
- The test picks its character from several scripts and skips when the ANSI codepage can represent all of them (or is UTF-8), so it is a no-op rather than a false failure under e.g. a Japanese locale.
- Ran `winmm:mmio`, `mci`, `wave`, `timer`, `mixer`. `mmio` goes FAIL → PASS; the other four are unchanged — `mci` has 73 pre-existing failures in this headless environment, identical with and without the patch.
- Also confirmed against a stock vs. patched Proton Experimental build (x86_64), same prefix layout and same test: stock fails with `0x101`, patched passes.
- **Verified in-game**: with a Proton Experimental build carrying only this patch, and no locale workaround, the game's sound effects work.
- Built and tested on the `proton_11.0` tree; `dlls/winmm/mmio.c` and `dlls/winmm/tests/mmio.c` are byte-identical between `proton_11.0` and `bleeding-edge`, so the change applies unmodified here.
- **Not verified against Windows**: I have no Windows machine. The test asserts the behaviour I believe native has (`mmioOpenW` being genuinely Unicode-capable for the default file IOProc). If native turns out to fail the same way, the new assertion should become `todo_wine` and this would be a bug-compatibility question rather than a fix.

Without this patch the only workaround is forcing an ANSI codepage that can encode the install path (`HOST_LC_ALL=ja_JP.UTF-8`), which is not something a user should have to discover.

Submitted upstream as https://gitlab.winehq.org/wine/wine/-/merge_requests/11606 — same two commits, cherry-picked onto wine master (the only difference between the two bases in these files is the Proton-only `#define WIN32_NO_STATUS` in `mmio.c`, which is nowhere near the change). If it lands upstream this PR can simply be closed in favour of the merge.

### DIFF
--- a/dlls/winmm/mmio.c
+++ b/dlls/winmm/mmio.c
@@ -54,8 +54,9 @@ static WINE_MMIO *MMIOList;
 /* From kernel32 */
 static HANDLE create_file_OF( LPCSTR path, INT mode )
 {
-    DWORD access, sharing, creation;
-    char full_path[MAX_PATH];
+    DWORD access, sharing, creation, len;
+    WCHAR *pathW, full_path[MAX_PATH];
+    HANDLE ret;
 
     if (mode & OF_CREATE)
     {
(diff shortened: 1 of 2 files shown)


kisak-valve:
Hello @mrq1911, yes, this should be reviewed and merged upstream first. After that's happened, than add a comment here with the upstream commit and a Proton dev can ponder a backport.

mrq1911:
Submitted upstream: https://gitlab.winehq.org/wine/wine/-/merge_requests/11606

Same two commits cherry-picked onto `wine/wine` master — `dlls/winmm/tests/mmio.c` is byte-identical between the two bases and `dlls/winmm/mmio.c` differs only by the Proton-only `#define WIN32_NO_STATUS` in the include block, so the change applies unmodified either way.

Worth noting for the open question in the description: the WineHQ pipeline includes `test-win10-21h2-32/64` jobs, so upstream CI will settle whether the new assertion holds on native Windows. If it fails there, the assertion should become `todo_wine` in both places.

PR #351, opened 2026-08-27
ws2_32: HACK: Delay a BattlEye control record for The Crew Motorfest.

PhialsBasement:
Fixes ValveSoftware/Proton#9119.

The Crew Motorfest answers BattlEye's scan query with a 20 fragment report, one fragment per 100ms tick, re-sent every ~7s until the server has all of them. The game has two send buffers for that stream and frees one on the peer's ack, so with a round trip above 200ms fragment 2 is dropped every round and the server sends `Query Timeout` after 210s. The game sends a 94 byte control record on the same stream ~500ms before each round; holding that one datagram so its ack lands in the tick before fragment 2 frees a buffer for it, and the report completes in two rounds.

The hook in `WS2_sendto` matches the 94 byte send to UDP port 4000, completes it to the caller immediately and issues it from a timer after a hold of 550 minus the measured RTT (RTT is timed against the 62 byte acks on the same socket). Gated on `SteamGameId == 2698940`; `WINE_TCM_REPORT_DELAY_MS` takes `auto`, a fixed number of milliseconds, or `0` to disable. Below 190ms RTT the game is unaffected and the hold is off. Above ~300ms the game's drop cadence changes and this cannot complete the report; a `FIXME` reports it.

Verified on a 220ms link with packet captures of the relay stream: the mechanism was first prototyped with a `tc netem` rule delaying only that datagram (4 of 4 sessions completed the report in two rounds), then the patch was run against a control with the hold disabled on the same night and relay hosts. The control ran all 31 rounds and was terminated at query + 210.0s; the patched sessions moved the control record to ~170ms before fragment 0 and completed in two rounds. Fragment 2 gets through by chance about 1 time in 7 (optimistic), so the captures were checked for that signature rather than survival alone.

Hope to get a review from @gofman

### DIFF
--- a/dlls/ws2_32/socket.c
+++ b/dlls/ws2_32/socket.c
@@ -974,6 +974,169 @@ static void WINAPI socket_apc( void *apc_user, IO_STATUS_BLOCK *io, ULONG reserv
     func( NtStatusToWSAError( io->Status ), io->Information, (OVERLAPPED *)io, 0 );
 }
 
+/* HACK: The Crew Motorfest drops BattlEye report fragment 2
+ * when RTT exceeds its 100ms send tick. Delaying the 94 byte
+ * control record sent before each round frees a buffer. */
+
+#define TCM_REPORT_PORT       4000
+#define TCM_REPORT_CTRL_SIZE  94
+#define TCM_REPORT_FRAG_MIN   1000
+#define TCM_REPORT_ACK_SIZE   62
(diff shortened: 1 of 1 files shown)


PhialsBasement:
Closing this one out; for anyone finding it later, the same effect can be had without any Proton changes using a tc rule that delays only that 94 byte datagram, and I will post the details in the issue.

PR #355, opened 2026-09-16
rsaenh: avoid null deref in update_hash_impl/finalize_hash_impl

BirknerAlex:
## Summary

`RSAENH_CPDeriveKey()` reuses the caller's `CALG_HMAC` hash object to pad short
hash output up to the target key length, calling `init_hash()`, `update_hash()`
and `finalize_hash()` on it directly. For `CALG_HMAC`, `finalize_hash()` ends by
zeroing `pCryptHash->hash` (including the `ltc_hash_descriptor` pointer), while
`dwState` is not necessarily reset back. A subsequent `CryptHashData()` call on
the same hash object can then reach `update_hash_impl()`/`finalize_hash_impl()`
with `hash->desc == NULL`, which dereferences it unconditionally and crashes.

Observed as a reproducible native access violation (read near NULL, offset
matching a field inside `ltc_hash_descriptor`) inside `CPHashData()`, in a game
that derives a session key from a running HMAC hash object during its login
handshake and continues hashing on that same object afterwards. Does not crash
on real Windows, so real RSAENH apparently tolerates this call pattern.

## Fix

Guard both `update_hash_impl()` and `finalize_hash_impl()` against a NULL
descriptor and fail gracefully (`FALSE`) instead of crashing.

## Test plan

- Built `dlls/rsaenh` for i386-windows and x86_64-windows against the
  `experimental_11.0` branch (commit at time of writing).
- Swapped the patched `rsaenh.dll` into a local Proton Experimental copy and
  reproduced the crash with the unpatched DLL, then confirmed it no longer
  occurs with the patch, using the actual game that triggered this (Aniimo,
  Steam appid 4126040) reaching well past the point that previously crashed
  every run.


kisak-valve:
Hello @BirknerAlex, you've come across a downstream fork of wine used in Proton. This merge request should be reviewed and merged into upstream wine first. After that's happened, then add a comment here with a link to the commit and a Proton dev can ponder a backport.

neptuwunium:
Patch does not apply to upstream wine as wine moved to SymCrypt in this commit: https://gitlab.winehq.org/wine/wine/-/commit/fc6d6ebbfe975059055062cfef312fc368bdff69

This bug, or at least the TomCrypt revision seems to be the regression that is causing https://github.com/ValveSoftware/Proton/issues/9427#issuecomment-5701985428

